PR #127, opened 2015-01-19
base: active header tab designation

### LINKED ISSUE #99: ui: show active tab
ui-show-active-tab

bouzlibop:
* Adds highlight to an active header tab. (closes #99)

### DIFF
--- a/zenodo/base/views.py
+++ b/zenodo/base/views.py
@@ -25,7 +25,7 @@
 import copy
 import time
 
-from flask import Blueprint, current_app, render_template
+from flask import Blueprint, current_app, render_template, request
 from flask.ext.breadcrumbs import register_breadcrumb
 from flask.ext.login import current_user
 from flask.ext.menu import current_menu, register_menu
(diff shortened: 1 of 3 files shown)


zenodo/base/views.py
```diff
@@ -55,6 +55,7 @@ def register_menu_items():
 
     item = current_menu.submenu('main.browse')
     item.register('', _('Browse'), order=3)
+    item._active_when = lambda: request.endpoint.startswith("search.collection")
```
bouzlibop: @lnielsen, could you take a look ? I think accessing the `private` attribute in this way is really bad practise. Is there any better way to do it?

lnielsen: You should be able to provide in  ``item.register(active_when=...)`` (see https://github.com/inveniosoftware/flask-menu/blob/master/flask_menu/__init__.py#L75)

bouzlibop: Sorry, that was a huge oversight :see_no_evil:

lnielsen:
Merge via 78e184f428b1889a5771bf9f15e3ff8cd696a979

PR #135, opened 2015-01-23
github: list your repositories GitHub API changes

### LINKED ISSUE #86: github: review organization permision changes
See: https://developer.github.com/changes/2014-12-08-organization-permissions-api-preview/

>Preview period:
"Staring __today__ [...]"

>Migration period:
"At the end of the preview period, we will announce the start of a migration period [...] We expect the migration period to last for four weeks [...]"

bouzlibop:
* moves from using `users/:username/repos` to `user/repos`
  API call to get the list of the users repositories. (closes #86)

* NOTE: List your repositories GitHub API now also includes
        repositories owned by organizations which the user
        can access.

        This functionality will apply to all API consumers
        beginning February 24, 2015.

### DIFF
--- a/zenodo/modules/github/utils.py
+++ b/zenodo/modules/github/utils.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 ##
 ## This file is part of Invenio.
-## Copyright (C) 2014 CERN.
+## Copyright (C) 2014, 2015 CERN.
 ##
 ## Invenio is free software; you can redistribute it and/or
 ## modify it under the terms of the GNU General Public License as
@@ -126,13 +126,17 @@ def sync(gh, extra_data, sync_hooks=True):
         depositions=[],
         errors=None,
     )
-    login = extra_data["login"]
 
     existing_set = set(extra_data['repos'].keys())
     new_set = set()
 
-    for u in [login] + [x.login for x in gh.iter_orgs()]:
-        for r in gh.iter_user_repos(u, type='owner', sort='full_name'):
+    repos = gh.iter_repos(type='all', sort='full_name')
+    repos.headers['Accept'] = 'application/vnd.github.moondragon+json'
+
+    for r in repos:
+        # Next line to be replaced with "r.permissions['admin']" once
+        # fix has been integrated in github3.py
+        if r.to_json()['permissions']['admin']:
             # Add if not in list
             if r.full_name not in extra_data["repos"]:
                 extra_data["repos"][r.full_name] = copy.copy(repo_data)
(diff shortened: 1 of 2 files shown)


bouzlibop:
@lnielsen, I have tested changes introduced here in my local environment. To be able to use the features from the upcoming API I switched the `Accept` filed in `github3` lib to the appropriate one (according to [this](https://developer.github.com/v3/repos/#list-your-repositories)), as so this PR __should not be__ applied until official migration period.

The results I got are like this:
* I was able to see the list of all my repositories (the one I own)
* I was also able to see the list of the repositories that are owned by the organizations I belong to (ie. eg. repos accesible for Invenio Developers Team, but no the ones which are restricted to Owners Team).

If this is what we need, this PR should solve the issue it is referring to.

lnielsen:
Just for reference in the future. The ``Accept``-header sent by github3.py can be modified like this:
```python
repos = gh.iter_repos(type='all', sort='full_name')
repos.headers['Accept'] = 'application/vnd.github.moondragon+json'
```

lnielsen:
There's a problem with ``gh.iter_repos(type='all', sort='full_name')``. It also returns repositories that you have no admin access to, and there seem to be no easy way to filter them. Similarly ``gh.iter_repos(type='owner', sort='full_name')`` only returns your own repositories and not the organisation repositories.

I've checked and the old approach is still valid to use (we just making more requests that is necessary).

lnielsen:
GitHub API does provide a list of permissions we can filter on:

```
    "permissions": {
      "admin": true,
      "push": true,
      "pull": true
    }
```

They are just not available through ``github3.py`` package it seems like.

bouzlibop:
What about hard way like: `r.to_json()['permissions']['admin']` ?
PS: Sorry for not testing it so well in the first place - I wasn't sure about the resulting repo's permissions.

lnielsen:
Great. Yes, that works. No worries about the testing. It's not that easy to test the GitHub integration.

bouzlibop:
:+1:

zenodo/modules/github/utils.py
```diff
@@ -126,25 +126,23 @@ def sync(gh, extra_data, sync_hooks=True):
...
-            # Update description
-            extra_data["repos"][r.full_name]["description"] = r.description
-            new_set.add(r.full_name)
-
-            # # TODO:
-            # if sync_hooks:
-            #     for h in r.iter_hooks():
-            #         if h.name == 'web' and 'url' in h.config:
-            #             h.config['url'] ==
+    for r in gh.iter_repos(type='all', sort='full_name'):
```
lnielsen: Can I get you to rebase and fix this line into:

```python
    repos = gh.iter_repos(type='all', sort='full_name')
    repos.headers['Accept'] = 'application/vnd.github.moondragon+json'

    for r in repos:
        # Next line to be replaced with "r.permissions['admin']" once
        # fix has been integrated in github3.py
        if r.to_json()['permissions']['admin']:
```

I've verified that it works, so then I'll be able to merge.

bouzlibop:
Sure, I'll do that now.

bouzlibop:
Alright, it's green now. 

<small>NOTE: I've just introduced the changes you mentioned about, but didn't check it in action.</small>

Issue #87, opened 2014-12-11
github: support flat style shields badges


lnielsen:
See badges/shields#176

PR #154, opened 2015-02-04
github: activate badge style support for DOI

bouzlibop:
* Adds support for `style` while generating a badge for
  specific DOI. (addresses #87).

### DIFF
--- a/zenodo/modules/github/views/badge.py
+++ b/zenodo/modules/github/views/badge.py
@@ -104,18 +104,25 @@ def doi_badge(doi):
 
     if pid is None:
         return abort(404)
-    return badge(doi)
+    style = request.args.get('style', None)
+
+    return badge(doi, style)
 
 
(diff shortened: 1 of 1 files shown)


zenodo/modules/github/views/badge.py
```diff
@@ -104,18 +104,23 @@ def doi_badge(doi):
...
+    return badge(doi, style)
 
 
 @blueprint.route("/<int:user_id>/<path:repository>.png", methods=["GET"])
 @ssl_required
 def index_old(user_id, repository):
     """Legacy support for old badge icons."""
-    return redirect(url_for('.index', user_id=user_id, repository=repository))
+    full_url = url_for('.index', user_id=user_id,
+                       repository=repository, **request.args)
```
lnielsen: ``request.args`` contains unvalidated data so it should not be passed blindly. Better just extract style, and pass it on.

PR #215, opened 2015-03-24
base: altmetric improvement

lnielsen:
* Changes altmetric widget to always load, but only show when data is
  present.

Signed-off-by: Lars Holm Nielsen <lars.holm.nielsen@cern.ch>


bouzlibop:
Possible solution to the problem connected to empty `wall` should be adding this rule:

```css
.altmetric-embed:empty {
  display: none;
}
```

support: http://caniuse.com/#search=%3Aempty

PR #235, opened 2015-04-13
Invenio version upgrade.

lnielsen:


requirements.master.txt
```diff
@@ -26,5 +26,5 @@
 -r base.requirements.txt
 
 ## Invenio and Zenodo
--e git+https://github.com/zenodo/invenio.git@zenodo-master#egg=Invenio
+-e git+https://github.com/zenodo/invenio.git@zenodo-master-rebase#egg=Invenio
```
lnielsen: Should not be integrated. Just checking before pushing Invenio branch.

PR #299, opened 2015-06-03
records: url prefix in alternate ids deletion

### LINKED ISSUE #262: records: alternate identifiers for urls should not prefix "url:"
See e.g. https://zenodo.org/record/17158

bouzlibop:
* Makes sure url schema prefix is not shown. (closes #262)


bouzlibop:
@lnielsen this is a possible solution (screen below presenting how it looks). Is it the way we want to solve it?

before:
bef1

after:
aft1

PR #307, opened 2015-06-11
deposit: api documentation

lnielsen:
* Adds API documentation for restricted access mode, as well as
  access_conditions field. (#closes 263)

Signed-off-by: Lars Holm Nielsen <lars.holm.nielsen@cern.ch>


lnielsen:
Missing:

- [ ] Add test case and properly validate possibilities to trick system by invalid combinations of access mode and conditions.

PR #337, opened 2015-07-01
info: about page

LibrarPotter:
Edits to About, policy, and FAQ pages.
More edits to FAQ page to come once new questions are added.

### DIFF
--- a/zenodo/base/templates/zenodo/faq.html
+++ b/zenodo/base/templates/zenodo/faq.html
@@ -48,7 +48,7 @@
         </p>
     </li>
     <li>
-        <h5>What does it cost?/Can we pay for Zenodo services? </h5>
+        <h5>What does it cost? /Can we pay for Zenodo services? </h5>
         <p>{{config.CFG_SITE_NAME}} is free for the long tail of Science. If you are interested in preserving a large amount of resource hungry research, please <a href="/contact">contact us</a>, and we will be happy to discuss expanding upload limits and our operation costs/capacity. Donations are always welcome.</p>
     </li>
     <li>
--- a/zenodo/base/templates/zenodo/faq.html
+++ b/zenodo/base/templates/zenodo/faq.html
@@ -32,7 +32,7 @@
     </li>
     <li>
         <h5>What can I upload?</h5>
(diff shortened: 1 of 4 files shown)


zenodo/base/templates/zenodo/faq.html
```diff
@@ -44,16 +40,16 @@
...
         </p>
     </li>
     <li>
-        <h5>What does it cost / Can we pay for Zenodo services? </h5>
-        <p>Zenodo is free for the long tail of Science. In order to offer services to the more resource hungry research, we will introduce a ceiling to the free slice and offer paid for slices above, according to the business model developed within the sustainability plan. If you can't wait but immediately want to explore these paid for options, please <a href="/contact">contact us</a> and we will look at interim measures with you.</p>
+        <h5>What does it cost? /Can we pay for Zenodo services? </h5>
```
lnielsen: I would keep it as-is for now, until the donation model has gone up. Then we can change it.

zenodo/base/templates/zenodo/faq.html
```diff
@@ -70,23 +66,23 @@
...
     </li>
     <li>
         <h5>Where does the name come from?</h5>
         <p>{{config.CFG_SITE_NAME}} is derived from <a href="http://en.wikipedia.org/wiki/Zenodotus">Zenodotus</a>, the first librarian of the Ancient Library of Alexandria and father of the first recorded use of metadata, a landmark in library history.</p>
     </li>
     <li>
-        <h5>How much storage do CERN have available?</h5>
-        <p>{{config.CFG_SITE_NAME}} is currently a drop in the ocean. CERN currently stores more than 100PB of physics data from the <a href="http://home.web.cern.ch/about/accelerators/large-hadron-collider">Large Hadron Collider (LHC)</a>, and produces roughly 25PB per year when the LHC is running.</p>
+        <h5>How much digital storage-space does CERN have available?</h5>
```
lnielsen: Not sure I would go with: ``digital storage-space``. At least it is not a normal term in this context. Also, we cannot use e.g. ``disk space`` since the space is both on disk and tape.

lnielsen:
Ping @otron. This needs to be integrated before you make the upgrade script.

lnielsen:
Merged via f46a6a65789983bc62a0c597e0a078bbbfcc6279

PR #366, opened 2015-07-20
base: move static pages to pages module

otron:


zenodo/base/upgrades/zenodo_2015_07_15_static_pages.py
```diff
@@ -102,7 +603,7 @@ <h4 id="restapi-res-">Quick start <small></small><span class="pull-right show-on
 </li>
 
 <li><p>Try to access the API:</p>
-<pre>>>> r = requests.get("{{config.CFG_SITE_SECURE_URL}}/api/deposit/depositions")
```
lnielsen: We need to find a way to deal with this.

To have maintainable pages, we need to be able to use variables I think, otherwise this page will not work nicely on e.g. the sandbox (i.e. it will say https://zenodo.org instead of https://sandbox.zenodo.org).

otron: Yeah, unfortunately template variables don't work in the content of pages module pages.
The content is passed to the template engine as a variable itself, which I guess handles it as a regular string variable.
Two possible solutions:

1. Get the template engine to read the content as a part of the template source. Or have it do two passes (the variables in the content would be identified and replaced on the second pass). This seems like an awful idea from a security perspective.
2. Write a filter that does the replacements (`s/{{ *var *}}/varvalue/` for all the variables in the config file). This'll be slower than (1), but safer.

PR #402, opened 2015-10-05
installation: minor installation and docs fix

kiryx:


kiryx:
I will do another PR on this from another branch

PR #403, opened 2015-10-07
deposit: depositions list unit test fix

kiryx:
* Fixes the problem with request context remaining on stack


bruno314:
:shipit:

PR #537, opened 2016-04-19
deposits: deposits form for zenodo

SamiHiltunen:
* Adds jsonschema for Zenodo deposits form and an angular schema form
  definition.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -193,6 +193,7 @@ setup(
         ],
         'invenio_jsonschemas.schemas': [
             'zenodo_records = zenodo.modules.records.jsonschemas',
+            'zenodo_deposit = zenodo.modules.deposit.jsonschemas',
         ],
         'invenio_search.mappings': [
             'records = zenodo.modules.records.mappings',
--- a/zenodo/config.py
+++ b/zenodo/config.py
@@ -102,6 +102,12 @@ DEPOSIT_CONTRIBUTOR_TYPES = [
     dict(label='Other', marc='oth', datacite='Other'),
 ]
 
+#: Default JSON Schema for deposit
+DEPOSIT_DEFAULT_JSONSCHEMA = 'zenodo_deposit/deposit-v1.0.0.json'
+
+#: Angular Schema Form for deposit
+DEPOSIT_DEFAULT_SCHEMAFORM = 'json/zenodo_deposit/deposit_form.json'
+
 # Formatter
 # =========
 #: List of allowed titles in badges.
--- a/zenodo/modules/deposit/jsonschemas/__init__.py
+++ b/zenodo/modules/deposit/jsonschemas/__init__.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Zenodo.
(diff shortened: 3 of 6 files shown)


setup.py
```diff
@@ -170,6 +170,7 @@
             'zenodo_frontpage = zenodo.modules.frontpage.views:blueprint',
             'zenodo_search_ui = zenodo.modules.search_ui.views:blueprint',
             'zenodo_theme = zenodo.modules.theme.views:blueprint',
+            'zenodo_deposits = zenodo.modules.deposits.views:blueprint',
```
krzysztof: I believe we should have been consistent with module names plurality, thus have ``invenio-deposits`` rather than ``invenio-deposit``, but for now it's deposit thus we should refer to "deposit" whenever we are talking about the module.

Hence it should be ``zenodo_deposit = zenodo.modules.deposit.views:blueprint`` (singular)

setup.py
```diff
@@ -192,6 +193,7 @@
         ],
         'invenio_jsonschemas.schemas': [
             'zenodo_records = zenodo.modules.records.jsonschemas',
+            'zenodo_deposits = zenodo.modules.deposits.jsonschemas',
```
krzysztof: Same as above.

zenodo/config.py
```diff
@@ -102,6 +102,12 @@ def _(x):
     dict(label='Other', marc='oth', datacite='Other'),
 ]
 
+#: Default JSON Schema for deposits
+DEPOSIT_DEFAULT_JSONSCHEMA = 'zenodo_deposit/deposit-v1.0.0.json'
+
+#: Angular Schema Form for deposits
+DEPOSIT_DEFAULT_SCHEMAFORM = 'json/zenodo_deposits/deposit_form.json'
```
krzysztof: ``json/zenodo_deposit/deposit_form.json``

zenodo/modules/deposits/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Zenodo; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Deposits package."""
```
krzysztof: """Deposit package."""

Also the name of the directory.

setup.py
```diff
@@ -170,6 +170,7 @@
             'zenodo_frontpage = zenodo.modules.frontpage.views:blueprint',
             'zenodo_search_ui = zenodo.modules.search_ui.views:blueprint',
             'zenodo_theme = zenodo.modules.theme.views:blueprint',
+            'zenodo_deposit = zenodo.modules.deposit.views:blueprint',
```
krzysztof: This entry already exists just above.

SamiHiltunen: Fixed 👍

PR #545, opened 2016-04-27
global: WIP deposit and record jsonschemas fixes

krzysztof:
* Changes Deposit and Record jsonschemas.

### DIFF
--- a/zenodo/modules/records/jsonschemas/records/record-v1.0.0.json
+++ b/zenodo/modules/records/jsonschemas/records/record-v1.0.0.json
@@ -176,10 +176,12 @@
       "type": "number"
     },
     "doi": {
+      "title": "Digital Object Identifier",
       "description": "Digital Object Identifier (DOI).",
       "type": "string"
     },
     "isbn": {
+      "title": "International Standard Book Number",
       "description": "International Standard Book Number (ISBN).",
       "type": "string"
     },
@@ -189,6 +191,7 @@
     },
     "resource_type": {
       "description": "Record resource type.",
+      "default": "publication",
       "type": "object",
       "additionalProperties": false,
       "properties": {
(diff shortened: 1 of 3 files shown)


zenodo/modules/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -463,15 +489,24 @@
...
-    "thesis_university": {
-      "description": "Awarding university in case of a thesis.",
-      "type": "string"
-    },
-    "thesis_supervisors": {
-      "description": "Supervisors of thesis in order of importance.",
-      "type": "array",
-      "items": {
-        "$ref": "#/definitions/person_or_org"
+    "thesis": {
```
krzysztof: The ``thesis_university`` and ``thesis_supervisors`` merged under ``thesis`` is the only important change. Is it OK?

zenodo/modules/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -189,6 +191,7 @@
     },
     "resource_type": {
       "description": "Record resource type.",
+      "default": "publication",
```
krzysztof: Is it OK to specify a ``default`` here?

zenodo/modules/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -264,6 +270,7 @@
     "access_right": {
       "description": "Access right for record.",
       "type": "string",
+      "default": "open",
```
krzysztof: As above. Specifying ``default`` can be useful for required fields and form generation. The form generation is more useful for deposit, but it shouldn't hurt?

krzysztof:
There's a lot of changes to deposit jsonschema and angular form definition but the most important are the record jsonschema changes as the previous two are resulting from it.

krzysztof:
Should the "references" term in record jsonschema have a list of strings rather than objects with a single string field ?
https://github.com/zenodo/zenodo/pull/545/files#diff-e89960279e476f82c825c2e2df57ddb0R379

PR #555, opened 2016-05-03
records: file permissions

SamiHiltunen:
* Adds permissions for files.


tests/unit/records/test_permissions.py
```diff
@@ -0,0 +1,111 @@
...
+    ('owner', AccessRight.EMBARGOED, 200),
+    ('owner', AccessRight.CLOSED, 200),
+    ('admin', AccessRight.OPEN, 200),
+    ('admin', AccessRight.EMBARGOED, 200),
+    ('admin', AccessRight.CLOSED, 200),
+])
+def test_file_permissions(app, db, test_object,  # fixtures
+                          user, access_right, expected):
+    """Test file permissions."""
+    with app.app_context():
```
lnielsen: This one is not needed, since the app fixture is already creating the app context (see https://github.com/SamiHiltunen/zenodo/blob/file-permissions/tests/unit/conftest.py#L70-L71)

SamiHiltunen: Fixed 👍

PR #564, opened 2016-05-12
views: file listing based on permissions

### LINKED ISSUE #423: migration: record file download
- [ ] Documents/Records-UI integration (access control, list files).
- [x] Template: Display files on record page.
- [x] Access control: Closed access files (owner, admin, shared record link), Embargoed (owner, admin, shared record link, after date).

SamiHiltunen:
* Lists files in record view only if the viewer has a permission to
  view them. (closes #423)

### DIFF
--- a/tests/unit/records/test_models.py
+++ b/tests/unit/records/test_models.py
@@ -111,6 +111,13 @@ def test_access_right_embargo():
         AccessRight.EMBARGOED, embargo_date=today+timedelta(days=1)) \
         == 'embargoed'
 
+    # Should work with strings as well
+    assert AccessRight.get(
+        AccessRight.EMBARGOED, embargo_date='1253-01-01') == AccessRight.OPEN
+    assert AccessRight.get(
+        AccessRight.EMBARGOED,
+        embargo_date=str(today+timedelta(days=1))) == AccessRight.EMBARGOED
+
 
 def test_object_type():
     """Test object type."""
--- a/zenodo/modules/records/models.py
+++ b/zenodo/modules/records/models.py
@@ -30,11 +30,14 @@ import json
 from datetime import datetime
 from os.path import dirname, join
 
+import arrow
+from elasticsearch_dsl.utils import AttrDict
 from flask import current_app
 from flask_babelex import gettext
 from invenio_search import current_search_client
 from invenio_search.api import RecordsSearch
 from jsonref import JsonRef
+from six import string_types
 from speaklater import make_lazy_gettext
 
 _ = make_lazy_gettext(lambda: gettext)
@@ -58,6 +61,24 @@ class AccessRight(object):
         (CLOSED, _('Closed Access')),
     )
 
+    _icon = {
+        OPEN: 'fa-unlock',
+        EMBARGOED: 'fa-warning',
+        RESTRICTED: 'fa-ban',
+        CLOSED: 'fa-lock',
+    }
+
+    _description = {
+        OPEN: _('Files are publicly accessible.'),
+        EMBARGOED: _('Files are currently under embargo but will be publicly '
+                     'accessible after {date}.'),
+        RESTRICTED: _('You may request access to the files in this upload, '
+                      'provided that you fulfil the conditions below. The '
+                      'decision whether to grant/deny access is solely under '
+                      'the responsibility of the record owner.'),
+        CLOSED: _('Files are not publicly accessible.'),
+    }
+
     _category = {
         OPEN: 'success',
         EMBARGOED: 'warning',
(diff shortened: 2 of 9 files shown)


zenodo/modules/frontpage/templates/zenodo_frontpage/index.html
```diff
@@ -44,7 +44,7 @@
         {%- set publication_date = r.publication_date|from_isodate -%}
         {%- set creation_date = r._created|from_isodatetime -%}
         {%- set embargo_date = r.embargo_date|from_isodate -%}
-        {%- set object_type = r.resource_type|objecttype %}
+        {%- set object_type = r.resource_type.to_dict()|objecttype %}
```
lnielsen: Do you know why this one is needed? It's probably better to fix in the ``objecttype`` filter or on ``ObjectType`` model.

SamiHiltunen: Elasticsearch-dsl uses an AttrDict which does have a get method. I added this here for now but the proper fix should most likely go into elasticsearch-dsl
https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/utils.py#L76

zenodo/modules/records/models.py
```diff
@@ -68,6 +88,9 @@ class AccessRight(object):
     @staticmethod
     def is_embargoed(embargo_date):
         """Test if date is still under embargo."""
+        embargo_date = embargo_date \
+            if not isinstance(embargo_date, string_types) \
+            else parser.parse(embargo_date).date()
```
lnielsen: For date parsing it's better to use arrow library for consistent among our modules (under the hood it uses dateutil):
```python
import arrow
arrow.get(value).date()
```

zenodo/modules/frontpage/templates/zenodo_frontpage/index.html
lnielsen: I agree that the fix is for es dsl - can you open an issue, but in the mean time move the logic to ``ObjectType`` model (e.g. the ``to_dict()`` i don't think will work if the record suddenly comes from the db).

zenodo/modules/records/models.py
```diff
@@ -68,7 +88,7 @@ class AccessRight(object):
     @staticmethod
     def is_embargoed(embargo_date):
         """Test if date is still under embargo."""
-        return embargo_date > datetime.utcnow().date()
+        return arrow.get(embargo_date).date() > datetime.utcnow().date()
```
lnielsen: Does this work with both ``datetime`` and ``string`` objects?

zenodo/modules/records/permissions.py
```diff
@@ -45,18 +45,30 @@ def __init__(self, bucket, action='objects-read'):
...
+            == AccessRight.OPEN:
+        return True
 
-        return False
+    user_id = int(user.get_id()) if user.is_authenticated else None
+
+    if user_id in record.get('owners', []):
+        return True
+
+    identity = Identity(user)
```
lnielsen: Is the ``UserNeed`` not automatically provided by Flask-Security on login? - https://github.com/mattupstate/flask-security/blob/develop/flask_security/core.py#L213

zenodo/modules/records/models.py
SamiHiltunen: This works with both datetime and string. I added a test to verify that.

zenodo/modules/records/permissions.py
SamiHiltunen: I loaded the identity so has_access could be called to verify user's access to a record without having to have the user currently logged in. I removed it now, should be easy to add it later if this is actually needed.

zenodo/modules/frontpage/templates/zenodo_frontpage/index.html
SamiHiltunen: I moved the check to ObjectType. Seems they removed the get method on purpose earlier. https://github.com/elastic/elasticsearch-dsl-py/commit/62ce0fc3684709e1370e00554b706eda53220075

tests/unit/records/test_models.py
```diff
@@ -111,6 +111,13 @@ def test_access_right_embargo():
         AccessRight.EMBARGOED, embargo_date=today+timedelta(days=1)) \
         == 'embargoed'
 
+    # Should work with strings as well
```
lnielsen: 👍

zenodo/modules/records/models.py
```diff
@@ -165,6 +196,8 @@ def get_by_dict(cls, value):
         if not value:
             return None
         if 'subtype' in value:
+            if isinstance(value, AttrDict):
```
lnielsen: 👍

lnielsen:
LGTM - just needs a rebase.

SamiHiltunen:
@lnielsen rebased

PR #569, opened 2016-05-18
global: zenodo-jsonschemas module

### LINKED ISSUE #575: tests: invalid test fixture
This test fixture initiates the Deposit API with invalid data:
https://github.com/zenodo/zenodo/blame/master/tests/unit/conftest.py#L471-L494

I guess this is supposed to be input to the REST API, not a deposit record, e.g. "metadata" is no longer a parameter in the new jsonschema.

krzysztof:
* Adds ZenodoJSONSchemas module for JSON schema compilation.

* Adds first version of compiled JSON schemas for record and deposit.

Closes #575

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -342,7 +342,6 @@ def full_record():
             'publisher': 'Jol'
         },
         communities=['zenodo'],
-        provisional_communities=['ecfunded'],
         grants=[
             {'title': 'Grant Title', 'code': '1234', 'identifiers': {},
              'internal_id': '10.1234/foo::1234',
@@ -472,17 +471,14 @@ def license_record(db):
 def deposit(app, es, users, location):
     """New deposit with files."""
     record = dict(
-        metadata=dict(
-            upload_type='presentation',
-            title='Test title',
-            creators=[
-                dict(name='Doe, John', affiliation='Atlantis'),
-                dict(name='Smith, Jane', affiliation='Atlantis')
-            ],
-            description='Test Description',
-            publication_date='2013-05-08',
-            access_right='open'
-        )
+        title='Test title',
+        creators=[
+            dict(name='Doe, John', affiliation='Atlantis'),
+            dict(name='Smith, Jane', affiliation='Atlantis')
+        ],
+        description='Test Description',
+        publication_date='2013-05-08',
+        access_right='open'
     )
     with app.test_request_context():
         datastore = app.extensions['security'].datastore
--- a/tests/unit/deposit/test_loaders_legacyjson.py
+++ b/tests/unit/deposit/test_loaders_legacyjson.py
@@ -315,8 +315,11 @@ def test_communities():
     )
 
 
-def test_prereserve_doi():
-    """Test communities."""
+def disabled_test_prereserve_doi():
+    """Test communities.
+
+    TODO: This test is disabled until DOI pre-reservation has been implemented.
+    """
     assert t(
         prereserve_doi=True,
     ) == defaults(_deposit_actions=dict(prereserve_doi=True))
--- a/zenodo/modules/deposit/jsonschemas/deposits/records/record-v1.0.0.json
+++ b/zenodo/modules/deposit/jsonschemas/deposits/records/record-v1.0.0.json
@@ -1,16 +1,839 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
-  "id": "http://zenodo.org/schemas/deposits/records/record-v1.0.0.json",
-  "title": "Zenodo Deposit Schema v1.0.0",
-  "allOf": [
-    {
-      "$ref": "https://zenodo.org/schemas/deposits/deposit-v1.0.0.json"
+  "additionalProperties": false,
+  "definitions": {
+    "contributor_person_or_org": {
+      "additionalProperties": false,
+      "properties": {
+        "affiliation": {
+          "description": "Affiliation for the purpose of this specific record.",
+          "type": "string"
+        },
+        "familyname": {
+          "description": "Family name of person if known.",
+          "type": "string"
+        },
+        "givennames": {
+          "description": "Given names of person if known.",
+          "type": "string"
+        },
+        "gnd": {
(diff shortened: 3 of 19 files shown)


krzysztof:
It would be nice to use OrderedDict for loading/dumping of source/destination schemas to preserve the order of keys as they are defined in JSON files, but that would require modification in invenio-jsonschemas as it uses ``json.load`` internally.

krzysztof:
Also, a unit test which checks if given "compiled" jsonschema is different than the one that's currently stored in the repository.

krzysztof:
Depends on https://github.com/inveniosoftware/invenio-jsonschemas/pull/28

This is so that OrderedDict can be used for schema loading, thus allowing to preserve the keys order within compiled jsonschema.

tests/unit/conftest.py
```diff
@@ -332,7 +332,6 @@ def full_record():
             'publisher': 'Jol'
         },
         communities=['zenodo'],
-        provisional_communities=['ecfunded'],
```
krzysztof: Because of ``"additionalParameters": false`` this is no longer allowed, but I guess it's only for ES?

krzysztof:
Depends on https://github.com/inveniosoftware/invenio-deposit/pull/68

tests/unit/conftest.py
```diff
@@ -472,17 +471,14 @@ def license_record(db):
...
-            access_right='open'
-        )
+        title='Test title',
+        creators=[
+            dict(name='Doe, John', affiliation='Atlantis'),
+            dict(name='Smith, Jane', affiliation='Atlantis')
+        ],
+        description='Test Description',
+        publication_date='2013-05-08',
+        access_right='open'
```
krzysztof: This was edited as this fixture was initiating Deposit with legacy metadata (this didn't pass the jsonschema validation)

tests/unit/deposit/test_loaders_legacyjson.py
```diff
@@ -315,8 +315,11 @@ def test_communities():
     )
 
 
-def test_prereserve_doi():
-    """Test communities."""
+def disabled_test_prereserve_doi():
```
lnielsen: Do we need to disable it?

zenodo/modules/deposit/loaders/schemas/legacyjson.py
```diff
@@ -199,7 +199,8 @@ class LegacyRecordSchemaV1(Schema):
     part_of = fields.Nested(PartOfSchema, attribute='metadata')
     imprint = fields.Nested(ImprintSchema, attribute='metadata')
     thesis = fields.Nested(ThesisSchema, attribute='metadata')
-    _deposit_actions = fields.Nested(ActionSchema, attribute='metadata')
+    # _deposit_actions = fields.Nested(ActionSchema, attribute='metadata')
+    # TODO: Disabled until pre-reserved DOIs are being handled
```
krzysztof: @lnielsen Is it OK to disable this for now ?

zenodo/modules/deposit/jsonschemas/deposits/records/record-v1.0.0.json
```diff
@@ -1,16 +1,839 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
+  "additionalProperties": false,
```
lnielsen: Is it possible to add some sort of warning in the description or similar that this is an auto-generated schema that should not be modified. I know we publish it so the message should be gentle, but important that developers don't make any edits directly in the file.

zenodo/modules/jsonschemas/cli.py
```diff
@@ -0,0 +1,107 @@
...
+"""CLI for ZenodoJSONSchemas."""
+
+from __future__ import absolute_import, print_function
+
+import json
+
+import click
+from flask_cli import with_appcontext
+from flask import current_app
+from .compilers import compile_deposit_jsonschema, compile_record_jsonschema, \
```
lnielsen: import sorting?

zenodo/modules/jsonschemas/compilers.py
```diff
@@ -0,0 +1,100 @@
...
+            sub_schema_url = sub_schema_ref['$ref']
+            sub_schema = resolve_schema_url(sub_schema_url)
+            yield sub_schema
+
+
+def _compile_common(schema):
+    """Compile common parts for record and deposit."""
+    id_ = schema['id']
+    title = schema['title']
+    # We need to iter 'allOf' manually because jsonresolver
```
krzysztof: Since I do not care about loaded-ordering, as I compile jsonschemas into a key-softed this is no longer true, but I'd not want to spend time fixing this right now as this solution works.

zenodo/modules/records/jsonschemas/records/base-v1.0.0.json
```diff
@@ -177,6 +177,7 @@
       "properties": {
         "type": {
           "description": "General resource type.",
+          "default": "publication",
```
krzysztof: ``default`` will be useful for angular form...

zenodo/modules/records/jsonschemas/records/base-v1.0.0.json
```diff
@@ -208,9 +209,11 @@
       "type": "string"
     },
     "keywords": {
+      "title": "Keywords",
```
krzysztof: And so will the "title"s

lnielsen:
LGTM - i'll merge

zenodo/modules/deposit/jsonschemas/deposits/records/record-v1.0.0.json
krzysztof: If they do, this unit test should catch it: https://github.com/zenodo/zenodo/pull/569/files#diff-a38230d8a36856aa1ee35822e1d6c9d3R32

zenodo/modules/jsonschemas/cli.py
krzysztof: isort didn't complain.. does it look wrong ?

PR #577, opened 2016-05-27
deposit: update provider pid_type

hachreak:
Signed-off-by: Leonardo Rossi <leonardo.r@cern.ch>


hachreak:
ping @lnielsen

zenodo/modules/deposit/providers.py
```diff
@@ -32,6 +32,8 @@
 class ZenodoDepositProvider(DepositProvider):
     """Deposit identifier provider."""
 
+    pid_type = 'dep'
```
lnielsen: Is this necessary.......fixes have just been merge.... it should be ``depid``

PR #582, opened 2016-05-30
setup: pin flask 0.10.1

krzysztof:
* Pins the flask version to 0.10.1

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -22,6 +22,8 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -30,7 +32,7 @@
 from docutils.utils import get_source_line
 
 
-def _warn_node(self, msg, node):
+def _warn_node(self, msg, node, *args, **kwargs):
```
krzysztof: This was suddenly necessary because of this line in sphinx: https://github.com/sphinx-doc/sphinx/blob/1fbc0c9a1611e343e1386daa01c6229ffeab7555/sphinx/environment.py#L917

Perhaps we're missing an error that didn't occur before.

PR #583, opened 2016-05-30
global: zenodo deposit api

krzysztof:
* Adds custom  Zenodo API for deposit.


docs/conf.py
```diff
@@ -30,7 +32,7 @@
 from docutils.utils import get_source_line
 
 
-def _warn_node(self, msg, node):
+def _warn_node(self, msg, node, *args, **kwargs):
```
lnielsen: latest commit in cookie cutter

PR #592, opened 2016-06-03
deposit: automatic accepting of owned communities

krzysztof:
* Adds automatic accepting of communites owned by any of deposit owners.


krzysztof:
As far as I can tell :shipit:

PR #593, opened 2016-06-03
deposit: addition of sip creation

krzysztof:
* Adds SIP creation for deposit publishing

### DIFF
--- a/zenodo/config.py
+++ b/zenodo/config.py
@@ -181,6 +181,17 @@ DEPOSIT_REST_ENDPOINTS = dict(
     ),
 )
 
+# SIPStore
+# ========
+#: Default JSON schema for the SIP agent
+SIPSTORE_DEFAULT_AGENT_JSONSCHEMA = 'sipstore/agent-webclient-v1.0.0.json'
+
+#: Enable the agent JSON schema
+SIPSTORE_AGENT_JSONSCHEMA_ENABLED = True
+
+#: Max length of SIPFile.filepath
+SIPSTORE_FILEPATH_MAX_LEN = 1000
+
 # Formatter
 # =========
 #: List of allowed titles in badges.
--- a/zenodo/modules/deposit/api.py
+++ b/zenodo/modules/deposit/api.py
@@ -27,9 +27,10 @@
 from __future__ import absolute_import
 
 from os.path import splitext
+from flask import current_app
 
+from invenio_deposit.api import Deposit as _Deposit
 from invenio_communities.models import Community, InclusionRequest
-from invenio_deposit.api import Deposit
 from invenio_records_files.api import FileObject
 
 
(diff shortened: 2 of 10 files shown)


krzysztof:
Depends on #592

zenodo/config.py
```diff
@@ -181,6 +181,17 @@ def _(x):
...
+# SIPStore
+# ========
+#: Default JSON schema for the SIP agent
+SIPSTORE_DEFAULT_AGENT_JSONSCHEMA = 'sipstore/agent-webclient-v1.0.0.json'
+
+#: Enable the agent JSON schema
+SIPSTORE_AGENT_JSONSCHEMA_ENABLED = True
+
+#: Enable the agent JSON schema
+SIPSTORE_FILEPATH_MAX_LEN = 1000
```
lnielsen: We should sync the value with the corresponding config value in Files-REST.

zenodo/modules/deposit/api.py
```diff
@@ -213,3 +218,39 @@ def _publish_edited(self):
...
+            if request.remote_addr:
+                ip_address = request.remote_addr
+            if current_user.email:
+                email = current_user.email
+        else:
+            return None
+        return dict(email=email, ip_address=ip_address)
+
+    def _create_sips(self, pid, record):
+        """Create the SIP information for the record."""
```
lnielsen: ``current_user.get_id()`` i think - flask-login

zenodo/modules/deposit/api.py
```diff
@@ -213,3 +218,39 @@ def _publish_edited(self):
         if not record['communities']:
             del record['communities']
         return record
+
+    @staticmethod
+    def _build_agent_info():
+        """Build the SIP agent info."""
+        email = ''
+        ip_address = ''
```
lnielsen: I guess this will not happen, but we can still log the IP address even for non-logged in user. If you are thinking about programatic API usage without a request, then you can also check with ``has_request_context()`` from flask.

zenodo/modules/deposit/api.py
```diff
@@ -213,3 +218,39 @@ def _publish_edited(self):
...
+            return None
+        return dict(email=email, ip_address=ip_address)
+
+    def _create_sips(self, pid, record):
+        """Create the SIP information for the record."""
+        user_id = None if current_user.is_anonymous else current_user.id
+        agent = self._build_agent_info()
+        sip = SIP.create('json', json.dumps(record.dumps()), user_id=user_id,
+                         agent=agent)
+        recsip = RecordSIP(sip_id=sip.id, pid_id=pid.id)
```
lnielsen: Perhaps we can move some of the code to ``RecordSIP`` or make a ``ZenodoRecordSIP`` which knowns how to make a SIP out of a Zenodo record. We can discuss IRL

krzysztof: Done like this: https://github.com/zenodo/zenodo/pull/593/files#diff-547d9e4f70f5e9978d6aaa9906d34b2bR38

zenodo/config.py
krzysztof: This value: https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/config.py#L65
has not been overwritten in Zenodo anywhere. I guess this one can stay like this for now ?

PR #638, opened 2016-06-23
doi: fix doi pid creation

krzysztof:
* Fixes the DOI PID creation for zenodo and external PIDs.


krzysztof:
WIP, I want to add autoadding of zenodo and ecfunded comumnities.

lnielsen:
Merge via #647

PR #641, opened 2016-06-27
tests: fixture of redirection test

JavierDelgadoFernandez:
* Fixes the random fail test occuring in the redirection test because
  of the querystring order.

Signed-off-by: Javier Delgado <javier.delgado.fernandez@cern.ch>


lnielsen:
Merged via #647

PR #642, opened 2016-06-27
ui: community curation buttons on record page

krzysztof:


krzysztof:
Access for admin is still missing. What is the current canonical way for checking if user is admin ?

zenodo/modules/records/views.py
```diff
@@ -251,3 +252,39 @@ def records_ui_export(pid, record, template=None):
         return render_template(
             template, pid=pid, record=record,
             data=data, format_title=formats[fmt]['title'])
+
+
+def _can_curate(community, user, record, removal=False):
```
krzysztof: This is hacky, perhaps a better suggestion ?

krzysztof: @lnielsen also, permission to admin should be granted.

zenodo/modules/records/templates/zenodo_records/box/curate.html
```diff
@@ -0,0 +1,100 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Zenodo; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+-#}
+<script type="text/javascript" src="https://code.jquery.com/jquery-2.2.4.min.js"></script>
```
lnielsen: jQuery should already have been loaded.

zenodo/modules/records/templates/zenodo_records/box/curate.html
```diff
@@ -0,0 +1,100 @@
...
+<script>
+$(document).ready(function(){
```
lnielsen: Please move to separate module that can be loaded via require.js
See e.g. https://github.com/zenodo/zenodo/blob/master/zenodo/modules/theme/templates/zenodo_theme/communities/curate.html#L31

zenodo/modules/records/templates/zenodo_records/box/curate.html
```diff
@@ -0,0 +1,100 @@
...
+            }
+          }
+        });
+    });
+});
+</script>
+
+  <dl>
+    <dt>{{_('Community curation')}}:</dt>
+    {% if pending_comms %}
```
lnielsen: General in this file:
``{%`` - > ``{%-``

Otherwise we get a lot of whitespace in the HTML.

lnielsen:
I'm somewhat concerned about the number of SQL queries. Before PR:

screen shot 2016-06-28 at 23 40 21

After PR:

screen shot 2016-06-28 at 23 39 33

Double the number of queries and ~1/3 slower page generation. It looks to be primarily due to ``DynamicPermission``.  Perhaps we can simply check once if user is admin instead of on every community. Can you have a look?

zenodo/modules/records/templates/zenodo_records/box/curate.html
```diff
@@ -0,0 +1,100 @@
...
+        </dd>
+      {% endfor %}
+    {% endif %}
+    {% if accepted_comms %}
+      <h5><i class="fa fa-group"></i> <strong>Communities</strong></h5>
+      {% for comm, can_curate in accepted_comms %}
+        <dd>
+        {% if can_curate %}
+        <span class="pull-right" id="span-comm-{{ comm.id }}">
+        <button class="btn-remove btn btn-xs"
```
lnielsen: add ``btn-info``

krzysztof: Done.

FYI: ``btn-info`` is the blue one, not the "3D" one: http://v4-alpha.getbootstrap.com/components/buttons/#examples

krzysztof:
I have optimized the permission code a little bit.

krzysztof:
Depends on #644 (which then depends on #648)

krzysztof:
Moved JS to the module.

@lnielsen If tests pass all comments have been addressed.

PR #648, opened 2016-06-29
global: Flask and Flask-CLI upgrade

krzysztof:


krzysztof:
Also downgrades the invenio-files-rest to commit 5243826da9183ce3a5841c6d7b45a8fe95c3c57e

PR #652, opened 2016-07-01
docker: uwsgi usage instead of gunicorn

hachreak:
Signed-off-by: Leonardo Rossi <leonardo.r@cern.ch>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,8 @@ RUN apt-get update \
     && find /usr/share/doc -depth -type f ! -name copyright -delete
 
 # Basic Python and Node.js tools
-RUN pip install --upgrade pip setuptools ipython gunicorn \
+RUN pip install --upgrade pip setuptools ipython \
+    && pip install uwsgi \
     && npm update \
     && npm install --silent -g node-sass clean-css uglify-js requirejs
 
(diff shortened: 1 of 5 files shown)


krzysztof:
Hey,

``requirements.developer.txt`` is no longer in the repo: https://github.com/zenodo/zenodo/pull/652/files#diff-3254677a7917c6c01f55212f86c57fbfR44

use ``requirements.txt`` instead.

hachreak:
I see... I'm trying to fix it :)

hachreak:
ping @lnielsen

uwsgi.ini
```diff
@@ -0,0 +1,9 @@
+[uwsgi]
```
lnielsen: Please move this file to ``docker/uwsgi/``

uwsgi.ini
```diff
@@ -0,0 +1,9 @@
...
+http = 0.0.0.0:5000
+wsgi-file = zenodo/wsgi.py
```
lnielsen: please use ``module`` - http://uwsgi-docs.readthedocs.io/en/latest/Options.html#module

zenodo/wsgi.py
```diff
@@ -26,7 +26,10 @@
 
 from __future__ import absolute_import, print_function
 
-from .factory import create_app
+from zenodo.factory import create_app
```
lnielsen: all changes here are not necessary

uwsgi.ini
hachreak: what should I do for virtualenv?

Issue #671, opened 2016-08-02
deposit: ui form styling


slint:
- [x] Actions bar in top + bottom. (Save, Delete) `S`
- [x] Resource type: remove description, check if all icons can be on one line. `M`
- [x] Missing field titles: e.g. books. `S`
- [x] Check required fields red-star mark. `S`
- [x] Check placeholder texts (e.g. authors). `S`
- [x] Check icons for all fields. (ui-select form elements have an unconfigurable template...) `S`
- [x] Check panels a) optional vs required b) open vs closed c) titles. `S`
- [x] Panels descriptions should go out of panel header and into body - e.g. communities. `M`
- [x] help text/description text in small. `S`
- [x] Remove field description in right column. `S`
- [x] Authors: remove help text. `S`
- [x] Authors: Match style of "Add author". `M`
- [x] Contributors: subfield on one line. `S`
- [x] Thesis is missing. `M`
- [x] Related identifiers: relations list should match reduced set from zenodo, hide scheme. `M`
- [x] Array/List fields: remove box around. `M`
- [x] Funder field - ~~funder + grant~~ - by default set to European Commission. `L`
- [x] Fix sortable lists (requires https://github.com/angular-ui/ui-sortable and sortOptions). `L`
- [x] Publication date/embargo date: date selector ([datepicker](https://github.com/Textalk/angular-schema-form-datepicker)) `M`
- [x] Select/Checkbox default values (can only be set in the deposit jsonschema). `M`
- [x] Add enhanced Markdown editor to Description ([ckeditor](https://github.com/webcanvas/angular-schema-form-ckeditor)) `L`
- [x] References: see if textarea field can be used, and process/split lines aferward. Use array list instead. `M`
- [x] License: placeholder text (click and type...), default to CC-BY-4.0, ~~if dataset CC0-1.0~~. `L`
- [ ] ~~Pre-reserve DOI button-click~~ Use ``{'metadata': {'prereserve_doi': {'doi' : ....}}}`` and add it to DOI on 'Save'.  `L`

**Misc enhancements**
- [x] Override/overwrite templates from other modules.
- [ ] Validation styling for all types of fields (green/tick, red/cross)

lnielsen:
Can you check alignment of the form as well. Probably just a ``<div class="container">`` missing somewhere:

Correct:

[...]

PR #703, opened 2016-08-11
deposit: enhanced form fields and legacy schema

slint:
* Configures autocomplete fields to their API endpoints and passes
  the necessary url parameters. (addresses #671)

* Adds custom templates for various field types.

* Adds datepicker form field type for dates.

* Adds ckeditor form field type.

* Adds ui-sortable for array fields that need ordering.

* Uses legacy record jsonschema for form validation.

* Adds app wide record serializer template filter.


slint:
Depends on inveniosoftware/invenio-deposit#111

PR #730, opened 2016-08-31
global: custom record_to_index mapping

krzysztof:


zenodo/config.py
```diff
@@ -867,3 +867,21 @@ def _(x):
...
 #: Number of proxies in front of application.
 WSGI_PROXIES = 0
+
+
+# Indexer
+# =======
+#: Provide a custom record_to_index function for invenio-indexer
+INDEXER_RECORD_TO_INDEX = "zenodo.modules.indexer.utils.record_to_index"
+INDEXER_SCHEMA_TO_INDEX_MAP = {
+    'http://zenodo.org/schemas/records/record-v1.0.0.json': ('records-record-v1.0.0', 'record-v1.0.0'),
```
lnielsen: Can't we reduce this to:

```python
{
   '<index name>': '<mapping name>',
  #...
}
```
and use schema_to_index, to provide us with the index name.....?

krzysztof: We can, but then our impl. would have to rely on the schema_to_index which I'd like to avoid.

For example, the original impl returns some defaults upon failure: https://github.com/inveniosoftware/invenio-indexer/blob/master/invenio_indexer/api.py#L65-L66 which is a 'silent' failure. Because this impl. raises when the schema is not mentioned specifically I have already found some records which were indexed incorrectly because they didn't have "$schema" in the JSON and were indexed to the "default".

krzysztof: I implemented the suggestion.

lnielsen:
Just an import sort failure in the failing tests....can you update

krzysztof:
Done

PR #748, opened 2016-09-07
global: pinned npm packages versions

krzysztof:


krzysztof:
@lnielsen FYI, this PR now comes with updated definition of bucket quota.

PR #750, opened 2016-09-07
WIP deposit: addition of custom data function

drjova:
this is related to https://github.com/inveniosoftware/invenio-records-js/pull/24

now each invenio app can override the `invenio-records-js` data preparation function.


lnielsen:
@slint @krzysztof is this one still needed or what is fixed in the mean time some other way?

slint:
I assume this was meant for keeping things compatible with `invenio-records-js > 0.0.6` (we have it pinned on v0.0.6 atm). I can attempt an upgrade instead and check if anything actually breaks...

lnielsen:
Ok , perhaps let's keep it on WIP for now, until we have to deal with JWT tokens and anyway need to upgrade.

slint:
Since we're already overriding some directives, it's ok to close this PR

PR #751, opened 2016-09-07
global: deposits, github and accounts fixes

slint:
Do not merge yet, depends on inveniosoftware/invenio-records-js#25, inveniosoftware/invenio-deposit, and inveniosoftware/invenio-accounts releases.


lnielsen:
Merged via #752

PR #834, opened 2016-09-26
marc21: addition of marc21 rule for conference URL

krzysztof:

### DIFF
--- a/zenodo/modules/utils/cli.py
+++ b/zenodo/modules/utils/cli.py
@@ -26,17 +26,22 @@
 
 from __future__ import absolute_import, print_function
 
+import os
+from io import SEEK_END, SEEK_SET
+
 import click
 from flask_cli import with_appcontext
-from invenio_pidstore.resolver import Resolver
-from invenio_records.api import Record
+from invenio_db import db
+from invenio_files_rest.models import ObjectVersion
 
 from zenodo.modules.deposit.tasks import datacite_register
+from zenodo.modules.records.api import record_resolver
 
 
 @click.group()
 def utils():
     """Zenodo helper CLI."""
+
 
 @utils.command('datacite_register')
 @click.argument('recid', type=str)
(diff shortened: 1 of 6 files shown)


krzysztof:
@lnielsen Sorry, I thought this was merged already and pushed the files CLI to my master (which is this PR)

zenodo/modules/utils/cli.py
```diff
@@ -26,28 +26,181 @@
...
+    """Add a new file to a publishd record."""
+    pid, record = record_resolver.resolve(recid)
+    bucket = record.files.bucket
+    key = os.path.basename(fp.name)
+
+    obj = ObjectVersion.get(bucket, key)
+    if obj is not None:
+        click.echo('File "{key}" alreay exists in the record {recid}'.format(
+            key=key, recid=recid))
+        return
```
lnielsen: The primary use case is to update an existing file. So perhaps add cli option ``-f/--force`` which will overwrite an existing file.

zenodo/modules/utils/cli.py
```diff
@@ -26,28 +26,181 @@
...
+
+    fp.seek(SEEK_SET, SEEK_END)
+    size = fp.tell()
```
lnielsen: I think ``fp.size`` should work so there's not need to seek in the file.

zenodo/modules/utils/cli.py
```diff
@@ -26,28 +26,181 @@
...
+    click.echo('Will add the following file:\n')
+    click.echo(click.style(
+        '  key: "{key}"\n'
+        '  bucket: {bucket}\n'
+        '  size: {size}\n'
+        ''.format(
+            key=key,
+            bucket=bucket.id,
+            size=size),
+        fg='red'))
```
lnielsen: ```red`` I would only use for actual errors. Here it's information.

zenodo/modules/utils/cli.py
```diff
@@ -26,28 +26,181 @@
...
+        click.echo('File "{key}" not found in the record {recid}'.format(
+            key=key, recid=recid))
+        return
+
+    if click.confirm('Will rename "{key}" to "{new_key}" on bucket {bucket}.'
+                     ' Continue?'.format(
+                        key=obj.key, new_key=new_key, bucket=bucket.id)):
+        record.files.bucket.locked = False
+
+        file_ = obj.file
```
lnielsen: ``file_id = obj.file_id``

zenodo/modules/utils/cli.py
```diff
@@ -26,28 +26,181 @@
...
+        ObjectVersion.delete(bucket, obj.key)
+        ObjectVersion.create(bucket, new_key, _file_id=file_.id)
```
lnielsen: ``..., _file_id=file_id``

zenodo/modules/utils/cli.py
krzysztof: No, as it an instance of io.BufferedReader

zenodo/modules/utils/cli.py
krzysztof: Changed to green

PR #837, opened 2016-09-28
UI and Task fixes

slint:


slint:
Stacked on #836, for merge convenience.

zenodo/modules/deposit/templates/zenodo_deposit/edit.html
```diff
@@ -82,6 +82,41 @@
           };
         });
 
+        // Overriding ui-select's controller to handle arrays
+        //
+        // Note: UI-Select by default doesn't play well while being inside an
+        // array. What happens is that if the array gets modified from the
+        // outside (eg. delete an array element), the changes are not picked up
+        // by UI-Select, and thus aren't refelcted on its internal model and
```
krzysztof: s/refelcted/reflected

zenodo/modules/records/views.py
```diff
@@ -253,9 +253,13 @@ def pid_url(identifier, scheme=None, url_scheme='https'):
...
         except IndexError:
             scheme = None
-    if scheme and identifier:
-        return idutils.to_url(identifier, scheme, url_scheme=url_scheme)
-    return ""
+    try:
+        if scheme and identifier:
+            return idutils.to_url(identifier, scheme, url_scheme=url_scheme)
+    except Exception:
+        current_app.logger.info('URL generation for identifier {0} failed.'
```
krzysztof: Perhaps a ``logger.warning``? Is this a common "error" that happens and should we ignore it, or do we want this in sentry?

slint: This originated from one specific error actually on Sentry, when there is a unicode character in some identifiers, specifically ['\u200b' zero-width-space](http://www.fileformat.info/info/unicode/char/200B/index.htm). We probably don't want to keep every malformed identifier error in sentry, since it's probably a user input issue...

krzysztof: :+1: Could we then catch ``UnicodeEncodeError`` here ? (I think that's the one that was being raised)

krzysztof:
This was merged already

PR #850, opened 2016-10-06
deposit: auto-requesting grant communities

krzysztof:
Signed-off-by: Krzysztof Nowak k.nowak@cern.ch


krzysztof:
With this PR communities should be auto-requested if they are specified in ``ZENODO_COMMUNITIES_REQUEST_IF_GRANTS`` config. ``ZENODO_COMMUNITIES_ADD_IF_GRANTS`` was left as is for future reference and should not be set in Zenodo.

krzysztof:
@lnielsen please review when you find a moment

PR #852, opened 2016-10-09
Fix grammar mistake on the sign up page

wetneb:
"Uploads gets a DOI" -> "Uploads get a DOI"


krzysztof:
@wetneb Thank you for the fix :)

Issue #865, opened 2016-10-24
bug in metadata subjects example


bneron:
on the web documentation API Documentation for developers ( https://zenodo.org/dev)
Resources > Representations > Deposition metadata > subjects

the example of json object for subject is :
[{"term": "Astronomy",
 "id": "http://id.loc.gov/authorities/subjects/sh85009003",
 "scheme": "url"}]

but id is not supported and the json is rejected 
the field must named 'identifier'

krzysztof:
Closed via #866 Thanks!

PR #866, opened 2016-10-24
fix issue #865

bneron:
replace field 'id' by 'identifier' in json example of subjects


krzysztof:
Indeed: https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/jsonschemas/records/record-v1.0.0.json#L500-L518

@bneron good catch, thanks :)

PR #868, opened 2016-10-26
global: UI fixes on records and deposit views

slint:


slint:
https://github.com/zenodo/zenodo/pull/868/files#diff-521b1bdf16088a18233cc3906647399cR101
This file has started to get a bit out of hand in terms of <script> tags... Since we have come across various ways to extend angular-schema-form, we should probably move all these "extensions" to a seperate module/JS-file/bundle.

slint:
A kind of major update of this PR is that now the "Prereserve DOI" button actually when pressed fires a request to the REST API in order to create the new deposit (in case the deposit that is being edited hasn't already been saved). This is done in a similar fashion as in Invenio-Files-JS, where uploading a file also triggers the creation/reservation of a depid/DOI.

krzysztof:
Sorry, you have to rebase :/

slint:
Rebased on #871

zenodo/modules/deposit/static/json/zenodo_deposit/deposit_form.json
```diff
@@ -287,6 +287,7 @@
       {
         "type": "textarea",
         "key": "notes",
+        "title": "Additional Notes",
```
lnielsen: We should normally strive to use UK english spelling i.e. lowercase n ``Additional notes``. (I know it's not consistently applied).

zenodo/modules/deposit/static/json/zenodo_deposit/deposit_form.json
```diff
@@ -503,7 +504,7 @@
       {
         "type": "array",
         "key": "related_identifiers",
-        "title": "Related identifiers",
+        "title": "Related Identifiers",
```
lnielsen: ditto here and the rest of the file.

PR #872, opened 2016-10-28
nginx: additional gzip mimetypes

slint:
- Updates nginx's gzip configuration with more mimetypes.

Signed-off-by: Alexander Ioannidis a.ioannidis@cern.ch


slint:
Currently JS files aren't being gzipped actually. The new list of mimetypes was pulled and merged with the old one from [this repository's sample nginx config](https://github.com/h5bp/server-configs-nginx/blob/master/nginx.conf#L92-L116).

PR #882, opened 2016-11-02
config: image preview max file size increase

lnielsen:


lnielsen:
Hopefully in the coming days but we have a critical issue with OAI-PMH harvesting which we're dedicating all resources to right now.

PR #884, opened 2016-11-03
deposit: oai syncing cli

krzysztof:
Signed-off-by: Krzysztof Nowak <k.nowak@cern.ch>

PR #887, opened 2016-11-07
deposit: published record fields preservation fix

krzysztof:
Fixes an issue with internal metadata fields such as _oai, _files getting overwritten after editing of a pre-migration record.


krzysztof:
Depends on #884

PR #893, opened 2016-11-11
utils: addition of files metadata repair CLI

krzysztof:


krzysztof:
This can be merge but should not be run uuid by uuid, instead of a batch task.

PR #894, opened 2016-11-14
github: addition of github CLI

krzysztof:

### DIFF
--- a/zenodo/modules/github/cli.py
+++ b/zenodo/modules/github/cli.py
@@ -1,0 +1,348 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2016 CERN.
+#
+# Invenio is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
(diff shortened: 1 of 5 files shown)


zenodo/modules/github/cli.py
```diff
@@ -0,0 +1,307 @@
...
+    """Management commands for GitHub."""
+
+
+def user_param(ctx, param, value):
+    """Resolve user from CLI input parameter (email or User ID)."""
+    if value is None:
+        return
+    if '@' in value:  # Try to resolve as email
+        try:
+            user = User.query.filter_by(email=value.strip()).one()
```
slint: Could it be kind of dangerous to use the email for identifying the user in this case? It might be that the email on Zenodo is not the same as the one used on GitHub, and this could lead to transferring/assigning repository ownership to the wrong person. Maybe always displaying a warning or doing an extra check for the user's email on GitHub (through `invenio_github.api.GitHubAPI.me`) might minimize these sort of risk or at least ring a bell.

zenodo/modules/github/cli.py
```diff
@@ -0,0 +1,348 @@
...
+@repo.command()
+@click.argument('user', callback=user_param)
+@click.option('--sync', '-s', default=False, is_flag=True,
+              help='Sync the repository prior to listing.')
+@click.option('with_all', '--all', '-a', default=False, is_flag=True,
+              help='List all repositories available.')
+@click.option('skip_email', '--skip-email-verification', '-E', is_flag=True,
+              help="Skip GitHub email verification.")
+@with_appcontext
+def list(user, sync, with_all, skip_email):
```
lnielsen: Please rename to avoid overload builtin function (command can still be named list).

krzysztof:
@lnielsen Done

PR #905, opened 2016-11-22
auditor: oai_id fix and minor refactoring

krzysztof:

### DIFF
--- a/zenodo/modules/auditor/oai.py
+++ b/zenodo/modules/auditor/oai.py
@@ -28,7 +28,6 @@ from __future__ import absolute_import, print_function, unicode_literals
 
 from itertools import chain
 
-import sqlalchemy as sa
 from elasticsearch_dsl import Q
 from flask import current_app
 from invenio_cache import current_cache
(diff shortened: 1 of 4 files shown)


krzysztof:
@slint In the end I'm not fetching uuids but OAI-IDs.

zenodo/modules/auditor/oai.py
```diff
@@ -71,16 +69,18 @@ def _build_oai_sets_from_db(self):
...
-                cast(PersistentIdentifier.pid_value, sa.Integer),
+                PersistentIdentifier.pid_value,
                 RecordMetadata.json)
         )
 
-        for recid, record in record_oai_sets:
+        for oai_id, record in record_oai_sets:
             for s in record.get('_oai', {}).get('sets', []):
-                self.cache.sadd('{}:{}'.format(self.cache_prefix, s), recid)
+                oai_control_number = oai_id.split(':')[-1]
```
slint: One caution on this one that just came in mind: We have OAI IDs like `zenodo.org:123`, but also like `cern.ch:123`. Since we only keep the last part of the identifier (the integer part) if eg. an OAI set contains both `zenodo.org:123` and `cern.ch:123`, our calculated set will only contain one element (`set([1])`), so we might end up with wrong results.

lnielsen:
@krzysztof Can this one be closed? I think you fixed it some other way?

PR #914, opened 2016-12-01
serialization: addition of files marcxml field

krzysztof:


krzysztof:
@lnielsen the leader field is missing in this PR, it also does not appear in the marcxml export in legacy, was this ever exported?

krzysztof:
Closes #914

zenodo/modules/records/serializers/to_marc21/rules.py
```diff
@@ -107,3 +107,18 @@ def reverse_conkerence_url(dummy_self, key, value):
...
+
+@to_marc21.over('856', '^_files$')
+@utils.reverse_for_each_value
+@utils.filter_values
+def reverse_files(dummy_self, key, value):
+    """Reverse - Files."""
+    return {
+        's': str(value['size']),
+        'u': value['uri'],
+        'w': value['checksum'],
```
lnielsen: In the old MARC export we didn't include the checksum and using [**$w - Record control number**](https://www.loc.gov/marc/bibliographic/bd856.html) I don't think would be correct.

Also, we did add **$z - Public note** with the ordering of the item, but I think we can leave it out.

krzysztof: Yes, I misinterpreted it - I'll remove the $w.
Should the checksum be in $z perhaps? I think checksum is more important to include than ordering?

krzysztof:
marcxml export: Adds leader field and files for open-access records.

krzysztof:
also needs communicating that $z in files will now return the md5 checksum

krzysztof:
@lnielsen Can you take another look?

PR #926, opened 2016-12-08
deposit: unicode sanitization

### LINKED ISSUE #917: OAI-PMH export failures due to non utf8 characters in text
Text in descriptions/titles/authors needs to be cleaned with ftfy tool on the deposit saving (add washing to marshmallow schema).

slint:
* Adds custom a custom marshmallow string field that washes strings
  using `ftfy` and a set of unwanted Unicode characters. (closes #917)

### DIFF
--- a/zenodo/modules/records/serializers/fields/sanitizedunicode.py
+++ b/zenodo/modules/records/serializers/fields/sanitizedunicode.py
@@ -22,25 +22,34 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
-"""Custom marshmallow fields."""
+"""Sanitized Unicode string field."""
 
 from __future__ import absolute_import, print_function
 
-from .datetime import DateString
-from .doi import DOI, DOILink
-from .html import SanitizedHTML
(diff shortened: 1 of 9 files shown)


zenodo/modules/records/serializers/fields/sanitizedunicode.py
```diff
@@ -0,0 +1,55 @@
...
+
+from ftfy import fix_text
+
+from .trimmedstring import TrimmedString
+
+
+class SanitizedUnicode(TrimmedString):
+    """String field that sanitizes and fixes problematic unicode characters."""
+
+    UNWANTED_CHARACTERS = {
```
krzysztof: 1. Are those the only problematic characters that can occur or just some examples of characters we had in the data that caused problems?
2. Does the ftfy solve the serialization issue without this explicit removal of characters?

slint: 1. All characters that I found on a run of the `marc` serializer on production records
2. Some of these characters weren't always solved/removed by `ftfy`, so including them here even if `ftfy` removes them is the only way to make sure they don't occur in the future.

Issue #828, opened 2016-09-23
indexer/search: add computed field for number of files


lnielsen:
Allow you to search for records with more than X number of files.

PR #952, opened 2017-01-18
records: subjects metadata field features

slint:


slint:
After merging this, recreation of the Elasticsearch `records` index is needed for searches on the subjects field to work.

lnielsen:
Ok, then we should probably pool the deployment together with #828

PR #955, opened 2017-01-25
global: file count/size indexing

lnielsen:


zenodo/modules/deposit/indexer.py
```diff
@@ -70,3 +70,8 @@ def indexer_receiver(sender, json=None, record=None, index=None,
     else:
         json['_updated'] = record.updated
     json['_created'] = record.created
+
+    # Compute filecount and total file size
+    files = json.get('_files', [])
+    json['filecount'] = len(files)
+    json['filesize'] = sum([f.get('size', 0) for f in files])
```
slint: `filesize` as a field name might give the impression of searching for the individual file size (Example query: `filesize:>1000000`), except of course if one is thinking of a record as a single file/archive. Alternatives: `record_size`, `total_filesize`, `dataset_size`, `archive_size`.

krzysztof: I second the name change, more alternatives: ``size``, ``filessize``

lnielsen: I agree that it is kind of misleading, the reasoning behind the name was to ensure that all the file related fields are predictable and simple.... ``filename, filetype, filesize, filecount``.

I'd prefer either ``size`` or ``bytesize``. WDYT?

slint: I vote for `size` out of these two. Another one: `datasize`

krzysztof: bytesize :-1:
size, datasize :+1:

PR #968, opened 2017-02-20
template: footer update with user docs links

krzysztof:


krzysztof:
This still needs old URL redirects

krzysztof:
This requires wiping all pages from Zenodo DB (``Pages.query.delete()``), as all have been replaced by a docs page.

Also we might want to update the initial blog post before the deployment.

krzysztof:
NOTE: When this is merged, also check if the readthedocs were built automatically.

krzysztof:
gsoc_news_zenodo

PR #973, opened 2017-02-27
Updated Readme.rst

jainaman224:
I have visited your page and found that:
1. Readme.rst : Add link to Install.rst 
>>> [Install.rst](https://github.com/zenodo/zenodo/blob/master/INSTALL.rst)

2. Readme.rst : Add gitter badge
>>> Added [Join the chat at https://gitter.im/zenodo/zenodo](https://gitter.im/zenodo/zenodo?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)

3. Readme.rst : Correct invenio link
>>> [Invenio](http://github.com/inveniosoftware/invenio)

These are some improvement according to me.

PR #975, opened 2017-02-28
INSTALL.rst : Correct the command to reindex the migrations

satwikkansal:
Change `zenodo migration reindex recid` to `zenodo migration reindex -t recid`


krzysztof:
@satwikkansal Thanks :+1: 

Can you please just change the commit title according with this comment: https://github.com/zenodo/zenodo/pull/973#issuecomment-283006445

so something like ``docs: correction in the reindexing command``

Issue #959, opened 2017-02-02
With apostrophe in title no citation is displayed


kwenzig:
There is a box for citations on the right, where an example for a generic citation is displayed, above the field where a citation style can be selected.

It seems that an apostrophe in the title (or author etc.) prevents this generic citation from being displayed.

Examples:
- https://zenodo.org/record/259550
- https://sandbox.zenodo.org/record/64230

PR #984, opened 2017-03-01
records: apostrophe in citation fix

satwikkansal:
Fixes https://github.com/zenodo/zenodo/issues/959

### DIFF
--- a/zenodo/modules/records/templates/zenodo_records/box/share.html
+++ b/zenodo/modules/records/templates/zenodo_records/box/share.html
@@ -36,6 +36,7 @@
   <h4>{{_('Cite as')}}</h4>
   <div id='invenio-csl'>
     {% set citation_result = record|citation(pid, style='apa') %}
+    {% set citation_result = citation_result.replace("'", "\\'").replace('"', '\\"') %}
     <invenio-csl
       {%- if citation_result %}
         ng-init="vm.citationResult = '{{citation_result}}'"


krzysztof:
This fix should not go in the template jinja code, but rather to the``citation`` template_filter which is defined here:
https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/views.py#L235-L245

Moreover, as you will see the template filter code doesn't do much, other than calling the serializer which is taking a record and a pid, and returns the citation (Line 240 in the file above), hence it's the serializer fault which is producing the citation without escaping the apostrophes properly.

The serializer in (``citeproc_v1``) is instantiated here: https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/serializers/__init__.py#L85

Hence, the fix to this should be applied to the serializer code (CiteprocSerializer in ``invenio-records-rest`` module here: https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/serializers/citeproc.py

Here's a guide on how to checkout local Invenio version of a module and test it with Zenodo: http://zenodo.readthedocs.io/en/latest/development/guide.html

satwikkansal:
Thank you @krzysztof for detailed description. Submitted a PR for the fix upstream at `invenio-records-rest`, so closing this PR :)
https://github.com/inveniosoftware/invenio-records-rest/pull/138

PR #988, opened 2017-03-04
Updated init.sh according to install.rst

rusrushal13:
I just changed the file as @slint said.


krzysztof:
Hey @rusrushal13 Actually it should stay in init.sh and be removed from the installation guide under "Demo records" section here: http://zenodo.readthedocs.io/en/latest/installation.html#demo-records (since the init.sh is run prior).

rusrushal13:
Okay I will update the installation guide, that will work fine I guess.

PR #990, opened 2017-03-06
Update INSTALL.rst

rajat1994:
@krzysztof ..... People are getting confused on how to change the path to virtualenvwrapper.sh. So, i put that part of the statements in the INSTALL.rst so that they don't need to look around.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -127,12 +127,13 @@ Elasticsearch, Redis and RabbitMQ:
     $ docker-compose up db es cache mq
 
 Keep the docker-compose session above alive and in a separate shell, create a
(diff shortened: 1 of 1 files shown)


krzysztof:
Unfortunately they'll have to look around.. :wink: 

In the linked URL to virtualenvwrapper, and on the very first page, there are very clear installation instructions of virtualenvwrapper. We have to assume that people read the basic documentation of linked dependencies and resources as they install them. The part that you included is a standard installation instructions of the virtualenvwrapper, so they didn't follow the installation guide fully if they haven't sourced the virtualenvwrapper.

Otherwise we should be consistent start copy pasting docker install docs, docker-compose install docs, etc... you see the problem ;)

Issue #974, opened 2017-02-28
global: test zenodo with elasticsearch v5.x


lnielsen:

satwikkansal:
@lnielsen I'd like to work on this issue. Should I integrate Travis into my fork or is it okay to create a PR and test on `zenodo` (origin) itself :)

krzysztof:
@satwikkansal better set up your own fork on travis, otherwise you'll be queued up a lot with zenodo PRs.

rusrushal13:
#1005

rusrushal13:
@lnielsen @krzysztof As es 2.x version doesn't support ingest attachment plugin so is it necessary to do include it?

rusrushal13:
@krzysztof @lnielsen @slint As installing V5.x is giving problems too, so can you suggest any other alternative in order to get it done.

rusrushal13:
@lnielsen is anyone is assigned to this issue?

lnielsen:
Now yes :-) I'm assigning @krzysztof since he's the one in charge of taking care of your PR.

rusrushal13:
@krzysztof is there I can help you with?

krzysztof:
Hey, thanks, but at the moment no - I'm not actively working on it. 
Since this issue proved to be difficult, it's probably not a low-hanging-fruit and requires longer investigation. We can't at the moment split it into smaller tasks since the efforts to to that might be equivalent to just solving it.. Nonetheless, your work was not for nothing and is very appreciated, as you gave insight into the complexity of this migration :)

PR #1005, opened 2017-03-09
test zenodo with elasticsearch v5.x #974

rusrushal13:
plugged in v5.2.2 es and green ticks on travis CI


.travis.yml
```diff
@@ -53,22 +53,28 @@ services:
...
+  - REQUIREMENTS=prod E2E="no" SQLALCHEMY_DATABASE_URI="sqlite:///zenodo.db" ES_URL="https://download.elastic.co/elasticsearch/release/org/elasticsearch/distribution/tar/elasticsearch/2.2.0/elasticsearch-2.2.0.tar.gz"
+  - REQUIREMENTS=prod E2E="yes" SQLALCHEMY_DATABASE_URI="postgresql+psycopg2://postgres@localhost:5432/zenodo" ES_URL="https://download.elastic.co/elasticsearch/release/org/elasticsearch/distribution/tar/elasticsearch/2.2.0/elasticsearch-2.2.0.tar.gz"
+
+jdk:
+  - "oraclejdk8"  
 
 python:
   - "2.7"
   - "3.5"
+  - "pypy"
```
lnielsen: We not going to run on pypy so you can remove this as it will add extra minutes to the build.

lnielsen:
If you look in the Travis build, then ES5 installation is failing with:

```console
$ /tmp/elasticsearch/bin/plugin install -b mapper-attachments
/home/travis/build.sh: line 62: /tmp/elasticsearch/bin/plugin: No such file or directory
The command "/tmp/elasticsearch/bin/plugin install -b mapper-attachments" failed and exited with 127 during 
```

There's two issues:

- The CLI ``plugin`` in 2.x was renamed to ``elasticsearch-plugin`` in 5.x
- The 2.x plugin ``mapper-attachments`` was deprecated in 5.x in favour of [``ingest-attachment``](https://www.elastic.co/guide/en/elasticsearch/plugins/5.2/ingest-attachment.html)

We can merge the PR without having the tests pass on ES5, but the above issues needs to be fixed so that 5.x is installed correctly.

slint:
Another issue (which unfortunately tests don't cover) is that the [Suggester API had some changes](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_50_suggester.html) that will affect auto-completion fields in the deposit form. I think adding tests for features that depend on Elasticsearch (eg. OAI-PMH) is a necessary step to be able to proceed with upgrading to v5.x.

rusrushal13:
As 2.x version doesn't support ingest attachment so is it necessary? @lnielsen @krzysztof

krzysztof:
@rusrushal13 The plubin in 2.x it's called "mapper-attachments", in 5.x - "ingest-attachment". Depending on ES version, only either should be installed.

rusrushal13:
@krzysztof So should I install mapper attachment plugin or not?

krzysztof:
@rusrushal13 For 5.x no, for 2.x yes.

krzysztof:
@rusrushal13 Sorry, I just realized that we cannot support both at the same time. In that case you don't need mapper attachment.

rusrushal13:
@krzysztof in that case we need to remove the V2.x environment

rusrushal13:
@krzysztof @lnielsen @slint As installing V5.x is giving problems too, so can you suggest any other alternative in order to get it done.

Issue #481, opened 2016-03-23
testing: test validation of record jsonschemas and ES mappings


krzysztof:
JSONSchemas for record metadata, as well as elasticsearch mappings should be tested.

The unit tests should simply create some rich examples for all record types used in Zenodo and make sure the validation is passing/not-passing where expected.

rajat1994:
@krzysztof ... i'd like to work on this issue..could you please tell which files do i need to look for?

krzysztof:
This involves first getting familiar with the Zenodo metadata model, which is described (and validated) using JSON Schema, here: https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/jsonschemas/records/record-v1.0.0.json

This model describes structure and some required fields for Record's metadata.

This schema is just for the record (*record* is sometimes referred to as *published* record on Zenodo). There are two other important schemas:
- Deposit (unpublished record): https://github.com/zenodo/zenodo/blob/master/zenodo/modules/deposit/jsonschemas/deposits/records/record-v1.0.0.json
- "Legacy" Deposit, which is an older version of Deposit schema (we need to keep this for old REST API compatibility): (https://github.com/zenodo/zenodo/blob/master/zenodo/modules/deposit/jsonschemas/deposits/records/legacyrecord.json)

The task is to:
1) Define some minimal/exhaustive test metadata for record (some are already in the conftest), deposit and legacy deposit according with the schema specification
2) Validate it according to their corresponding schemas

Having unit tests like these protects from introducing "breaking" changes to the schema w.r.t. structure and "required" fields.

For start there are already two "minimal" and "full" test metadata fixtures:

https://github.com/zenodo/zenodo/blob/master/tests/unit/conftest.py#L316-L329
https://github.com/zenodo/zenodo/blob/master/tests/unit/conftest.py#L406-L548

And there are even tests which test the "minimal" metadata constraints:
https://github.com/zenodo/zenodo/blob/master/tests/unit/records/test_schemas.py#L36-L38

At first it would be good to add similar validation for full_record fixture.

rajat1994:
@krzysztof ..... i have validated some of the parameters of Deposit schema. Could you please check?
Link of gist :  https://gist.github.com/rajat1994/be2dea187460b2ed621419d55ca7bae4

krzysztof:
Hey @rajat1994,

That would be the general idea for task, however:

1) You don't have to define your own schema - Zenodo already has some defined as I mentioned above.
[...]

PR #1011, opened 2017-03-10
updated test_schemas.py

rajat1994:
i have added some tests for validation of  the full_record fixture similar to minimal.

### DIFF
--- a/tests/unit/records/test_schemas.py
+++ b/tests/unit/records/test_schemas.py
@@ -42,14 +42,16 @@ def test_recid(app, db, minimal_record):
     """Test recid property."""
     # String instead of number
     minimal_record['recid'] = '123'
-    pytest.raises(ValidationError, Record.create, minimal_record)
+    exc = pytest.raises(ValidationError, Record.create, minimal_record)
+    assert exc.value.message == "'publication' is not of type 'object'"
 
 
 def test_resource_type(app, db, minimal_record):
     """Test resource type."""
     # String instead of number
     minimal_record['resource_type'] = 'publication'
-    pytest.raises(ValidationError, Record.create, minimal_record)
+    exc = pytest.raises(ValidationError, Record.create, minimal_record)
+    assert exc.value.message == "'publication' is not of type 'object'"
     minimal_record['resource_type'] = {'type': 'publication', 'subtype': 'x'}
     Record.create(minimal_record)
 
@@ -71,8 +73,8 @@ def test_contributors(app, db, minimal_record):
     minimal_record['contributors'] = [
         {'name': 'test', 'affiliation': 'test', 'type': 'Invalid'}
     ]
-    pytest.raises(ValidationError, Record.create, minimal_record)
-# validation of full_record fixture
+    exc = pytest.raises(ValidationError, Record.create, minimal_record)
+    assert exc.value.message == "'publication' is not of type 'object'"
 
 
 def test_full_json(app, db, full_record):
@@ -80,29 +82,33 @@ def test_full_json(app, db, full_record):
     Record.create(full_record)
 
 
-def full_test_recid(app, db, full_record):
+def test_full_recid(app, db, full_record):
     """Test recid property."""
     full_record['recid'] = '12345'
-    pytest.raises(ValidationError, Record.create, full_record)
+    exc = pytest.raises(ValidationError, Record.create, full_record)
+    assert exc.value.message == "'publication' is not of type 'object'"
 
 
-def full_test_resource_type(app, db, full_record):
+def test_full_resource_type(app, db, full_record):
     """Test resource type."""
     full_record['resource_type'] = 'publication'
-    pytest.raises(ValidationError, Record.create, full_record)
+    exc = pytest.raises(ValidationError, Record.create, full_record)
+    assert exc.value.message == "'publication' is not of type 'object'"
     full_record['resource_type'] = {'type': 'publication', 'subtype': 'book'}
     Record.create(full_record)
 
 
-def full_test_contributors(app, db, full_record):
+def test_full_contributors(app, db, full_record):
     """Test contributors."""
     full_record['contributors'] = [
-       {'affiliation': 'CERN', 'name': 'Smith, Other', 'type': 'Other',
+       {'affiliation': 'CERN', 'name': ' '', '' ', 'type': 'Other',
                        'gnd': '', 'orcid': '0000-0002-1825-0097'}
     ]
-    Record.create(full_record)
+    exc = pytest.raises(ValidationError, Record.create, full_record)
+    assert exc.value.message == "'publication' is not of type 'object'"
     full_record['contributors'] = [
-        {'affiliation': '', 'name': 'Hansen, Viggo', 'type': 'Other',
+        {'affiliation': '', 'name': 'Hansen, Viggo', 'type': 'Author',
                             'gnd': '', 'orcid': ''}
     ]
-    pytest.raises(ValidationError, Record.create, full_record)
+    exc = pytest.raises(ValidationError, Record.create, full_record)
+    assert exc.value.message == "'publication' is not of type 'object'"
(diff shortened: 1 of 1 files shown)


rajat1994:
@krzysztof .... could you please review this?

tests/unit/records/test_schemas.py
```diff
@@ -72,3 +72,37 @@ def test_contributors(app, db, minimal_record):
...
     pytest.raises(ValidationError, Record.create, minimal_record)
+# validation of full_record fixture
+
+
+def test_full_json(app, db, full_record):
+    """Test full json."""
+    Record.create(full_record)
+
+
+def full_test_recid(app, db, full_record):
```
krzysztof: tests in pytest must start with "test", otherwise they are not executed.

tests/unit/records/test_schemas.py
```diff
@@ -72,3 +72,37 @@ def test_contributors(app, db, minimal_record):
...
+    """Test recid property."""
+    full_record['recid'] = '12345'
+    pytest.raises(ValidationError, Record.create, full_record)
+
+
+def full_test_resource_type(app, db, full_record):
```
krzysztof: Must start with "test_"

tests/unit/records/test_schemas.py
```diff
@@ -72,3 +72,37 @@ def test_contributors(app, db, minimal_record):
...
+    """Test resource type."""
+    full_record['resource_type'] = 'publication'
+    pytest.raises(ValidationError, Record.create, full_record)
+    full_record['resource_type'] = {'type': 'publication', 'subtype': 'book'}
+    Record.create(full_record)
+
+
+def full_test_contributors(app, db, full_record):
```
krzysztof: Must start with "test_"

tests/unit/records/test_schemas.py
```diff
@@ -72,3 +72,37 @@ def test_contributors(app, db, minimal_record):
...
+def full_test_recid(app, db, full_record):
+    """Test recid property."""
+    full_record['recid'] = '12345'
+    pytest.raises(ValidationError, Record.create, full_record)
+
+
+def full_test_resource_type(app, db, full_record):
+    """Test resource type."""
+    full_record['resource_type'] = 'publication'
+    pytest.raises(ValidationError, Record.create, full_record)
```
krzysztof: I know this was not in other tests, but let's introduce the following convention for this type of testing:

```
exc = pytest.raises(ValidationError, Record.create, full_record)
assert exc.value.message == "'publication' is not of type 'object'"
```
i.e. every time you check for validation error raising, additionally assert that the cause is as expected.

tests/unit/records/test_schemas.py
```diff
@@ -72,3 +72,37 @@ def test_contributors(app, db, minimal_record):
...
+    full_record['contributors'] = [
+       {'affiliation': 'CERN', 'name': 'Smith, Other', 'type': 'Other',
+                       'gnd': '', 'orcid': '0000-0002-1825-0097'}
+    ]
+    Record.create(full_record)
+    full_record['contributors'] = [
+        {'affiliation': '', 'name': 'Hansen, Viggo', 'type': 'Other',
+                            'gnd': '', 'orcid': ''}
+    ]
+    pytest.raises(ValidationError, Record.create, full_record)
```
krzysztof: This will not fail because the input is valid. See the specification of the "contributors": https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/jsonschemas/records/record-v1.0.0.json

Try "type" that's not in an enum. Then below in the same test try skipping the "name" field (with a proper "enum").

tests/unit/records/test_schemas.py
```diff
@@ -72,3 +72,37 @@ def test_contributors(app, db, minimal_record):
         {'name': 'test', 'affiliation': 'test', 'type': 'Invalid'}
     ]
     pytest.raises(ValidationError, Record.create, minimal_record)
+# validation of full_record fixture
```
krzysztof: Remove the comment.

krzysztof:
Addresses https://github.com/zenodo/zenodo/issues/481

tests/unit/records/test_schemas.py
rajat1994: @krzysztof ... i'm getting assertion error

tests/unit/records/test_schemas.py
```diff
@@ -71,4 +73,42 @@ def test_contributors(app, db, minimal_record):
...
+
+
+def test_full_contributors(app, db, full_record):
+    """Test contributors."""
+    full_record['contributors'] = [
+       {'affiliation': 'CERN', 'name': ' '', '' ', 'type': 'Other',
+                       'gnd': '', 'orcid': '0000-0002-1825-0097'}
+    ]
+    exc = pytest.raises(ValidationError, Record.create, full_record)
+    assert exc.value.message == "'publication' is not of type 'object'"
```
krzysztof: ``'publication' is not of type 'object'`` was an example for the first test, each test will raise different error. Please ipdb here and see what the actual error for each test.

Also, I noticed that this text will format differently between Py2 and Py3:
https://travis-ci.org/zenodo/zenodo/jobs/210554938#L5366

Can you check if you can get some Py2/Py3-independent message from this exception?

rajat1994: @krzysztof .... i'm getting the "IOError: reading from stdin while output is captured"  while using ipdb

krzysztof: You have to run the pytest with '-s' : ``pytest tests/test_file.py -s``

rajat1994: @krzysztof .... i debugged the test_recid() method  and it is  giving the following exception 
https://gist.github.com/rajat1994/6e2457067c602ebb31478b5bb9e7340a

krzysztof: I commented on your gist.

rajat1994: @krzysztof .... i have pasted the assertion error in test_resource_type. Similar  assertion errors i'm getting in all other test cases. Could you please check the gist?

krzysztof: The error:
```
E assert "'publication...ype u'object'" == "'publication'...type 'object'"
E - 'publication' is not of type u'object'
E ? -
E + 'publication' is not of type 'object'
```
 happens because the formatting of this message differs between Python 2 and Python 3, because of the differences between unicode handling (notice the extra 'u' before the 'object'). To fix this one, you need to see if you can somehow render this validation error message, independent of the Python version.

rajat1994:
@krzysztof ... could you please take a look at it?... i have handled the assertion error.

tests/unit/records/test_schemas.py
```diff
@@ -38,31 +38,33 @@ def test_minimal_json(app, db, minimal_record):
...
 
 
-def test_recid(app, minimal_record):
+def test_recid(app, db, minimal_record):
     """Test recid property."""
     # String instead of number
     minimal_record['recid'] = '123'
-    pytest.raises(ValidationError, Record.create, minimal_record)
+    exc = pytest.raises(ValidationError, Record.create, minimal_record)
+    assert (exc.value.message)
```
krzysztof: This only checks if the exception message (python string) evaluates to True, which is always True for all non-empty strings ;)

The idea was to check if the content of message is accurate e.g. that 'recid' cannot be a string but a number etc.
Then in a test below 'resource type' has to be an object and not just a string. etc.

If you can't make it work due to Py2/Py3 dependencies, just remove those checks completely, i.e. just check if pytest.raises as before.

krzysztof: This only checks if the exception message (python string) evaluates to True, which is always True for all non-empty strings ;)

The idea was to check if the content of message is accurate e.g. that 'recid' cannot be a string but a number etc.
Then in a test below 'resource type' has to be an object and not just a string. etc.

If you can't make it work due to Py2/Py3 dependencies, just remove those checks completely, i.e. just check if pytest.raises as before.

tests/unit/records/test_schemas.py
```diff
@@ -38,31 +38,33 @@ def test_minimal_json(app, db, minimal_record):
...
     minimal_record['resource_type'] = {'type': 'publication', 'subtype': 'x'}
     Record.create(minimal_record)
 
 
 def test_publication_date(app, db, minimal_record):
-    """Test recid property."""
-    # String instead of number
+    """Test publication date."""
+    # String instead of numbe
```
jainaman224: typo numbe should be number

PR #1013, opened 2017-03-13
global: removing pages fixtures

krzysztof:


lnielsen:
I've rebased this PR to latest master. Please check if we still need it.

PR #1022, opened 2017-03-17
footer menu update (about + status)

rusrushal13:


krzysztof:
This PR looks gook, but it needs to be synchronized with a corresponding change in user docs.
For that reason let's keep it open as a reminder and merge just before next deployment.

rusrushal13:
Okay 🙂

krzysztof:
Integrated and extended here: https://github.com/zenodo/zenodo/pull/1242

Thanks for the contribution :)

PR #1026, opened 2017-03-22
selenium: Add test_basic_tests.py and modify test_frontpage.py

kishansag:
Added test_basic_tests.py(Contains tests for login, registration, forgot password, etc)

Modified test_frontpage.py

The code of new file test_basic_tests.py is borrowed from https://github.com/inveniosoftware/invenio-accounts/blob/master/tests/e2e/e2e_basic_test.py


rajat1994:
@kishansag ... use ipdb to debug your code

PR #1032, opened 2017-03-25
records-ui: add a "Contact the owner." button to the record page

kishansag:
Have added the button "Contact the Owner". Only UI is done. Button doesn't send a message yet. I will add other functionalities quickly.
Uses bootstrap dialogbox, form would pop up on the same page as shown in here  http://getbootstrap.com/javascript/#live-demo


krzysztof:
Needs some more work to get to a finished state. Closing for now. Thanks a lot for the contribution though :)

PR #1008, opened 2017-03-10
global: contact form

jainaman224:
* Form is added for contact.
* Contact form automatically detects OS and browser.
* Sends information in the form of email notification.

Enhancement for #901.

PR #1034, opened 2017-03-25
Bug fix #1010

### LINKED ISSUE #1010: records-ui: add a simple "contact the author" button with contact form to the owner of the record.
Currently it's not easy to contact the owner of the record because we don't expose user emails.
We can however allow for contacting them through us and send them an email from the users.

It should be possible to opt-out of this (bool in the userprofile model).

vedantrathore:
@krzysztof sorry for the delay, I'll complete this in a day now :sweat_smile: I have added the template for the submission form, I am currently facing some errors on how to write the python code for email part, any suggestions would be highly appreciated.

### DIFF
--- a/zenodo/modules/records/views.py
+++ b/zenodo/modules/records/views.py
@@ -44,6 +44,8 @@ from werkzeug.utils import import_string
 from .models import AccessRight, ObjectType
 from .permissions import RecordPermission
 from .serializers import citeproc_v1
+from .forms import contact_form_factory
+
 
 blueprint = Blueprint(
     'zenodo_records',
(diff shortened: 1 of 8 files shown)


vedantrathore:
@krzysztof I've added some code for sending email, but not sure on how to proceed forward, could you please help :)

zenodo/modules/records/views.py
```diff
@@ -53,6 +55,43 @@
...
+
+    """If form is validated send email to the admin."""
+    if form.validate_on_submit():
+        """Dictionary storing data to be sent."""
+        context = dict(
+            form=form,
+            current_user=current_user,
+        )
+
+        # How to implement recipient user email?
```
vedantrathore: @krzysztof How do I add the recipient's (i.e. the person whose record is being investigated here) email address here? is there a method for getting user email from record PID?

vedantrathore: @krzysztof How do I add the recipient's (i.e. the person whose record is being investigated here) email address here? is there a method for getting user email from record PID?

vedantrathore:
I've added the code but not tested it yet, since it's under development. I will write the unit test after the feature is done

lnielsen:
Closed via #1008 #1239 and #1243

PR #1037, opened 2017-03-27
Bug fix #997

### LINKED ISSUE #997: serializers: BibTex escape special characters
Current BibTex export is not escaping special characters such as ampersands ("&"), causing trouble in applications that use LaTeX. Example record: https://zenodo.org/record/345946/export/hx.

A list/sepcification of these special characters should be found and escaped during serialization using backslashes (eg. `{ Some subject \& another subject }` ). More information can be found in [this SO thread](http://tex.stackexchange.com/questions/7198/how-to-escape-the-ampersand-character-in-bibtex).

vedantrathore:
@slint I've added the test case and the method, I've also validated the bibtex syntax

### DIFF
--- a/tests/unit/records/test_bibtex_serializer.py
+++ b/tests/unit/records/test_bibtex_serializer.py
@@ -206,27 +206,29 @@ def test_get_volume(bibtex_records):
     assert test_record['journal']['volume'] == record_good._get_volume()
     assert "" == record_empty._get_volume()
 
+
 def test_clean_input(full_record):
-  """Test."""
-  full_record['resource_type']['subtype'] = 'article'
-  full_record['title'] = "Test title & escaped character"
-  bibtex = ("""@article{doe_2014_12345,\n"""
-            """  author       = {Doe, John and\n"""
-            """                  Doe, Jane and\n"""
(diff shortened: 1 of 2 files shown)


vedantrathore:
@slint I dunno why this is failing, all the tests are passing on my pc

slint:
PEP8: https://travis-ci.org/zenodo/zenodo/jobs/215728929#L5340
You should run tests locally before pushing to Travis using the `./run-tests.sh` script on the root of the repository.

I also recommend to check out for any available Python linter integration for your editor such as `pyflake`, `pylinter`, etc. This way you can identify such issues quickly while developing and save time.

slint:
@vedantrathore Travis is still failing because of PEP8 errors. You should be running the tests using the `run-tests.sh` script, at least by calling `py.test --pep8 tests`. Usually error messages are easier to read locally.

vedantrathore:
@slint It's done :)

krzysztof:
Closing since this PR has been integrated as part of big PR cleanup here: https://github.com/zenodo/zenodo/pull/1241

Thanks a lot for the contribution and sorry for the delay :)

Issue #901, opened 2016-11-20
global: contact form


lnielsen:
Add a contact form to allow for easier support management:

- Form should ask people to include direct links and be precise. See https://github.com/support for example.
- Contact form can automatically detect which browser and OS they are using.

jainaman224:
Can I try to add this enhancement...?

krzysztof:
@jainaman224 Yes, that would actually be of huge help to our support :)

The main idea is that many people contact us (zenodo.org - the service), about common problems, which are usually:
- Removing a record, updating a file to a record somewhere
- Doing something with their account
- Getting some error on a page

Very often they don't include much info (such as the mentioned OS, browser), and very often they don't specify which record they want to remove/update etc, and sometimes they forget to mention what is their Zenodo email (and very often it's different than the one they're writing from).

The main idea would be to include as much of that info as possible and send it to us through the form, so

1. Capture basic agent info (browser, OS, IP). And send it with the email.
2. Capture the user if he's logged in and send his email + user ID in the email (form should recommend to log-in first).
3. Text-description specifying on what information they should include:
 - Record URL or DOI if it's relevant to the problem.
 - Error ID if they get one from a page (we can easily check it on Sentry)
 - URL of a page that might be crashing for them
 - GItHub username/orgname/reponame if it's GitHub related.

The task is to implement the points 1. and 2., we can think of a nice text on the form (point 3.) together later.

krzysztof:
This form should probably send us an email, where bottom part of the email body can include a pre-formatted string with this information.

jainaman224:
Do I need to implement this form https://zenodo.org/support ...?

krzysztof:
The current "/support" page is just static info without any form - it should stay like this for now.

You would need to create a new page (under e.g.: "/support-form"), where you would define some simple contact form similar to GitHubs: https://github.com/support

Here's an example of a form in Zenodo (this particular one is for deleting records by admin):

Template:
https://github.com/zenodo/zenodo/blob/master/zenodo/modules/deposit/templates/zenodo_deposit/delete.html

Form:
https://github.com/zenodo/zenodo/blob/master/zenodo/modules/deposit/forms.py

View:
https://github.com/zenodo/zenodo/blob/master/zenodo/modules/deposit/views.py#L142-L211

jainaman224:
Thank you :+1: .

krzysztof:
[...]

PR #1041, opened 2017-03-29
global: contact form

jainaman224:
- [x] python linter issues
- [x] multiple attachments
- [x] maximum upload size

Enhancement for #901.


jainaman224:
@krzysztof Do I need to bother about convention that there should be 1 import per line and no relative paths are allowed. As I have seen these conventions are not followed till date.

jainaman224:
Suggestions: Making text area auto expandable.

krzysztof:
> @krzysztof Do I need to bother about convention that there should be 1 import per line and no relative paths are allowed. As I have seen these conventions are not followed till date.

As long as isort does not complain it's OK. Relative paths have been allowed so far..

krzysztof:
Tests are failing because of isort actually. You can fix it automatically by running:
``isort -rc`` in zenodo repo (this will modify the code).

jainaman224:
I tried that way earlier but flake8 and isort has contradicting suggestion.
When I changes imports order according to isort. I get
`H306 — imports not in alphabetical order (flask_wtf.flaskform, flask_wtf.file.filefield)`	

Please help me out.

jainaman224:
:TODO: If the textarea expansion gets approved. Make generic js module and add jquery support for zenodo like mathjax.

zenodo/modules/pages/utils.py
```diff
@@ -81,9 +85,14 @@ def send_support_email(context):
...
     )
 
-    if context['form'].attachments.data:
-        msg.attach(context['form'].attachments.data.filename,
-                   'application/octet-stream',
-                   context['form'].attachments.data.read())
+    attachments = request.files.getlist("attachments")
+    if attachments:
+        for upload in attachments:
+            if len(upload.read()) > \
```
slint: This reads the whole file first (which might be eg. 5GB) and then rejects. There are other ways to prevent this from happening beforehand (@jainaman224 don't worry with resolving this for the time being, this is just a note)

zenodo/modules/pages/templates/zenodo_pages/contact_form.html
```diff
@@ -80,6 +80,8 @@
 {%- endblock page_body %}
 
 {%- block javascript %}
+<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.0/jquery.min.js"></script>
```
slint: You can skip this, since jQuery is by default loaded from the `zenodo.js` bundle (included through [require.js](https://github.com/zenodo/zenodo/blob/master/zenodo/modules/theme/bundles.py#L62-L63) from [this file](https://github.com/zenodo/zenodo/blob/master/zenodo/modules/theme/static/js/zenodo.js#L23))

jainaman224: Ok :+1:

jainaman224: Done and thanks. We need to include zenodo.js explicitly using `{% assets "zenodo_theme_js" %}<script src="{{ ASSET_URL }}"></script>{% endassets %}` .

requirements.txt
```diff
@@ -217,6 +217,7 @@ SQLAlchemy-Continuum==1.2.4
 SQLAlchemy-Utils==0.32.11
 statsd==3.2.1
 traitlets==4.3.1
+ua-parser==0.7.3
```
lnielsen: This requirement also has to be added in setup.py with  a ``>=`` instead of ``==``.

jainaman224: What is the convention about adding dependencies in `setup.py`? As I am unable to find some dependencies from `requirements.txt` in setup.py.

Thanks

lnielsen: Have a look at https://zenodo.readthedocs.io/en/latest/development/process.html#python-requirements and the link in that section. First a requirement should go in setup.py (abstract requirement), then a specific version goes in requirements.txt, and requirements.txt list every single package that Zenodo needs (pip freeze) so that we're sure that we don't actually break stuff by suddenly having a third party package upgrade.

jainaman224: :+1:

zenodo/modules/pages/views.py
```diff
@@ -51,39 +50,62 @@
 )
 def support():
     """Render contact form."""
-    form_class = contact_form_factory()
-    form = form_class()
+    form = ContactForm()
+    os, browser, device = user_agent_information()
```
slint: There's no need to unpack this tuple, since you always pass these parameters around as is. Instead of that you can  keep them in a `dict` and pass it.

zenodo/modules/pages/views.py
```diff
@@ -51,39 +50,62 @@
...
+            flash(
+                _('Request sent successfully, '
+                  'You should receive a confirmation email within 20 minutes -'
+                  ' if this does not happen you should retry or send us an '
+                  'email directly to team@zenodo.org.'),
+                category='success'
+            )
+            return redirect('/')
 
     content = current_app.config['PAGES_ISSUE_CATEGORY']
```
slint: You can move this line above, to keep everything related to the "view setup" together.

zenodo/modules/pages/views.py
```diff
@@ -51,39 +50,62 @@
...
-            _('Request sent successfully,'
-              'You should receive a confirmation email within 20 minutes - '
-              'if this does not happen you should retry or send us an email '
-              'directly to team@zenodo.org.',),
-            category='success'
-        )
-        return redirect('/')
+        if send_support_email(context) == "error":
+            flash(
+                _('File size exceeded, Please add url to file instead.'),
```
slint: Replace `,` with `.`

zenodo/modules/pages/views.py
```diff
@@ -51,39 +50,62 @@
...
+            )
+        else:
+            flash(
+                _('Request sent successfully, '
+                  'You should receive a confirmation email within 20 minutes -'
+                  ' if this does not happen you should retry or send us an '
+                  'email directly to team@zenodo.org.'),
+                category='success'
+            )
+            return redirect('/')
```
slint: ...or `url_for('zenodo_frontpage.index')`

zenodo/modules/pages/utils.py
```diff
@@ -81,9 +112,70 @@ def send_support_email(context):
...
     mail.send(msg)
+
+
+def user_agent_information():
+    """Function for user agent information."""
+    uap = user_agent_parser.Parse(str(request.user_agent))
+
+    # Operating System and it's version.
+    os = ''
+    if uap['os']['family'] is not None:
```
slint: You can simplify all this string building code using something like:
```python
os_version = '.'.join([v for v in (os['major'], os['minor'], os['patch']) if v])
'{os.family} {os_version}'.format(os=uap['os'], os_version=os_version)
```
You can also try to make it more generic in a small internal function maybe.

zenodo/modules/pages/utils.py
```diff
@@ -81,9 +112,70 @@ def send_support_email(context):
...
+        device = uap['device']['family']
+        if uap['device']['brand'] is not None:
+            device += ' ' + uap['device']['brand']
+        if uap['device']['model'] is not None:
+            device += ' ' + uap['device']['model']
+
+    return os, browser, device
+
+
+def display_size(size):
```
slint: You can also use the [`do_filesizeformat` Jinja function](https://github.com/pallets/jinja/blob/d3ed6c7947172fe96342a38c5fd779f205cf1527/jinja2/filters.py#L371-L398)

zenodo/modules/pages/utils.py
```diff
@@ -66,6 +70,33 @@ def format_request_email_body(context):
...
     return render_template_to_string(template, context)
 
+
+def check_attachment_size(attachments):
+    """Function to check size of attachment within limits.
+
+    :param attachments: Attachments files.
+    :returns: Boolean.
+    """
+    if check_attachment_size(attachments) == 0:
```
slint: I think you meant `len(attachments)` here...

zenodo/modules/pages/utils.py
```diff
@@ -66,6 +70,33 @@ def format_request_email_body(context):
...
+        return False
+    if attachments:
```
slint: ...although this is equivalent to `if len(attachments) != 0:`. So you can skip the above check and add it to the `else` clause

zenodo/modules/pages/utils.py
```diff
@@ -66,6 +70,33 @@ def format_request_email_body(context):
...
+        if 'Content-Length' in request.headers:
```
slint: There's also `request.content_length`. Using these properties of `request` offer sometimes some safety and "free" parsing

zenodo/modules/pages/utils.py
```diff
@@ -66,6 +70,33 @@ def format_request_email_body(context):
...
+        else:
+            size = 0
+            for upload in attachments:
+                upload.seek(current_app.config['PAGES_ATTACHMENT_MAX_SIZE'] -
+                            size)
+                if upload.read(1) == '':
+                    return False
+
+                upload.seek(0)
+                size += upload.read()
```
slint: I believe calling `read` at this point reads the actual bytes of the file and doesn't return the number of bytes. You are probably looking for something like `upload.seek(0); size += len(upload)`. You should test anyways with some large files to see what happens.

zenodo/modules/pages/templates/zenodo_pages/contact_form.html
```diff
@@ -24,63 +24,206 @@
...
+          {%- if each == 'include_os_browser' %}
+          <div class="row">
+            <div class="col-md-12">
+            <div class="col-md-2"></div>
+            <div class="col-md-10">
+              <input checked="" id="include_os_browser"
+                name="include_os_browser" type="checkbox" value="y"> Send the
+                browser and system information.
+              <div id="information">
+                {% if uap.os != '' %}
```
slint: Just `{% if uap.os %}` will do here (also in the next `if`s)

zenodo/modules/pages/templates/zenodo_pages/contact_form.html
```diff
@@ -24,63 +24,206 @@
...
+          {% else %}
+          <div class="alert alert-info login-email" role="alert">
+            Zenodo staff will get back to you by your Zenodo email:
+            <b>{{ form.email.data }}</b>.
+          </div>
+          {% endif %}
+
+          {{ form.csrf_token }}
+
+          {%- for each in form.field_sets %}
```
slint: `field` is a probably more descriptive name than `each` in the context of this loop

zenodo/modules/pages/templates/zenodo_pages/contact_form.html
```diff
@@ -93,11 +236,96 @@
     var dropdown = document.getElementById('issue_category');
     var current_value = dropdown.options[dropdown.selectedIndex].value;
 
-    for(each in data){
-      if(data[each][0]==current_value){
+    for(each in data)
```
slint: You can easily transform the `data` to a dict/object before and access the text by index

zenodo/modules/pages/utils.py
```diff
@@ -81,9 +110,47 @@ def send_support_email(context):
...
+        else:
+            return "error"
 
     mail.send(msg)
+
+
+def os_browser(info):
+    """Function to return string."""
+    info_version = '.'.join(
+        [v for v in (info['major'], info['minor'], info['patch']) if v]
```
slint: All these fields might be missing from the dict. You shold use `info.get('major')` instead of direct index access, especially since this information comes from a third-party library

zenodo/modules/pages/templates/zenodo_pages/contact_form.html
jainaman224: What about `form_field` as `field` is used in next line and that variable is used that way in mos modules. What is your opinion?

zenodo/modules/pages/views.py
```diff
@@ -51,39 +52,54 @@
...
 
-    """If form is validated send email to the admin."""
+    # Generation of message based upon maximum attachment size.
+    content = dict((a, c) for a, b, c in
+                   current_app.config['PAGES_ISSUE_CATEGORY'])
+
+    # Information added to form from config file.
+    form.issue_category.choices = [(row[0], row[1]) for row in
+                                   current_app.config['PAGES_ISSUE_CATEGORY']]
+    form.description.validators = [Length(
```
slint: This replaces existing validators. You should rather `form.description.validators.append(...)`

zenodo/modules/pages/utils.py
```diff
@@ -66,6 +70,31 @@ def format_request_email_body(context):
...
+            upload.seek(current_app.config['PAGES_ATTACHMENT_MAX_SIZE'] -
+                        size)
+            if upload.read(1) == '':
+                return False
+
+            size += len(upload)
+    return True
+
+
 def send_support_email(context):
```
slint: You should get the mail instance through the registered extension of `current_app`, as done [here](https://github.com/inveniosoftware/invenio-mail/blob/master/invenio_mail/tasks.py#L47)

zenodo/modules/pages/views.py
```diff
@@ -51,39 +52,54 @@
...
-
-        flash(
-            _('Request sent successfully,'
-              'You should receive a confirmation email within 20 minutes - '
-              'if this does not happen you should retry or send us an email '
-              'directly to team@zenodo.org.',),
-            category='success'
-        )
-        return redirect('/')
+        if send_support_email(context) == 'error':
```
slint: Retuning error codes is usually not a good practice (especially since this case is simple). Instead of that you can raise an exception inside the `send_support_email` function, or even bettter just call `check_attachment_size` here.

zenodo/modules/pages/views.py
```diff
@@ -51,39 +52,54 @@
...
+                category='success'
+            )
+            return redirect(url_for('zenodo_frontpage.index'))
 
-    content = current_app.config['PAGES_ISSUE_CATEGORY']
     return render_template(
         'zenodo_pages/contact_form.html',
+        uap=uap,
         form=form,
         current_user=current_user,
```
slint: `current_user` is alread part of the default template context (injected by `flask-security`), so this context argument is redundant.

zenodo/modules/pages/views.py
```diff
@@ -51,39 +52,54 @@
...
+        do_filesizeformat(current_app.config['PAGES_ATTACHMENT_MAX_SIZE'])
+
+    # If form is validated send email to the admin.
     if form.validate_on_submit():
-        """Dictionary storing data to be sent."""
+        # Dictionary storing data to be sent.
         context = dict(
+            uap=uap,
             form=form,
             current_user=current_user,
```
slint: `current_user` can always be accessed through `flask_security.current_user`, so it's better not to pass it around, but use the import.

zenodo/modules/pages/templates/zenodo_pages/contact_form.html
```diff
@@ -93,11 +218,96 @@
...
+        string += '<br /><small>' + files[i].name + ' &nbsp&nbsp ' +
+        display_size(files[i].size) + '</small>';
+      }
+
+      document.getElementById('upload_info').innerHTML = string;
+
+      if(totalSize > {{ max_file_size }}){
+        $('.field-attachments').addClass('has-error');
+        document.getElementById('error-attachments').innerHTML =
+        'File size exceeded. Please add URL to file instead.';
```
slint: Better text: `File size exceeded. Please add URLs to the files or make a smaller selection.`

zenodo/modules/pages/templates/zenodo_pages/contact_form.html
```diff
@@ -93,11 +218,96 @@
     var dropdown = document.getElementById('issue_category');
     var current_value = dropdown.options[dropdown.selectedIndex].value;
 
-    for(each in data){
-      if(data[each][0]==current_value){
-        document.getElementById('issue_text').innerHTML = data[each][1];
+    for(each in data)
```
slint: You can do `dropdown.options[dropdown.selectedindex].innerHTML = data[current_value]` here. 

Also as a side note, try to generally use intuitive and context-aware variable names inside your `for` loops (eg. `each in data` -> `description in descriptions`).

lnielsen:
@slint Just review with @krzysztof and you can merge.

Issue #942, opened 2017-01-13
deposit: removing DOI when editing a published record shouldn't be possible


slint:
Currently editing a published record and completely removing/clearing the DOI field in the deposit form, is a "Save"-able state, from which though the user cannot recover (clicking "Pre-reserve DOI", or filling in the DOI manually and saving results to an "internal server error").

In order to tackle this issue, two steps are required:
- [x] Proper validation of the submitted deposit data with regard to the fact that it's a deposit of an already published record.
- [ ] Disabling the DOI field in deposit form when the published record uses a Zenodo DOI.

krzysztof:
Steps to reproduce this error:

1. Create a new deposit, fill in all metadata + file, and provide an external DOI, e.g.: "10.1000/12345"
2. Click "Save" and "Publish"
3. Edit the record, remove the DOI, click "Save" (Saves the deposit properly)
4. At this point Saving the deposit throws Internal Server Error on Saving/Publishing.

kapilkd13:
Hi @krzysztof  I have reproduced the issue.
I have some questions regarding the step 1 you mentioned "Proper validation of the submitted deposit data with regard to the fact that it's a deposit of an already published record."
should we completely disallow users to genrate DOI if record is already published.
Also can you explain little  bit about Validation that need to be done

krzysztof:
The rules for DOI editing are as follows:

When you first create a deposit, you can either:
a) provide you own DOI which you might already gotten from a publisher
b) opt-in for a Zenodo DOI either by leaving the DOI field empty, or clicking pre-reserve DOI, which is simply showing you what your DOI will be (end result is, that the DOI is 'zenodo' DOI which we generate for the user).

Now, if you click publish for the first time, and if you chose a), you can later edit the DOI field (e.g. you might have made a typo in the DOI field - we wouldn't know that since we didn't register it ourselves). Editing of the DOI in case of a) is possible.

In case of b), it's no longer possible to change the DOI after you're published the record. When you open the deposit for editing the DOI field should be disabled (you can no longer change the DOI).

The DOI validation code is stored in the DOI field schema here: https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/serializers/fields/doi.py

krzysztof:
So in the problem description above at Step 3, it should not be possible to save an "empty" DOI, once the record has been published.

Second issue is that it should not be possible to Save an already published deposit (during editing), when the Zenodo DOI has been already registered.

[...]

PR #1043, opened 2017-04-01
WIP: disable removing of DOI on published record while editing

guptavaibhav18197:
Proper validation has been added 
to prevent user from removing DOI while editing a published record. 
ADDRESSING issue #942


guptavaibhav18197:
@krzysztof Sir
The user was able to leave the doi blank(even for a published record) because 
the metadata dictionary in data  [here](https://github.com/zenodo/zenodo/blob/master/zenodo/modules/deposit/loaders/base.py#L54) was not having any key named 'doi', so the validation doi validation function [here](https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/serializers/fields/doi.py#L78) was NOT being called.
In the PR I made a check - if the record is published and not having 'doi' key  -- then I will add a key,value pair - (doi,'') to the metadata dictionary and made appropriate validation check in the above mentioned validation function.
But the above is failing on [test](https://travis-ci.org/zenodo/zenodo/jobs/217412937#L5045) regarding editing drafts.
Can you please let me know the reason?

rajat1994:
@guptavaibhav18197 .... there are some assertion errors ... 
https://travis-ci.org/zenodo/zenodo/jobs/217412937....check this

guptavaibhav18197:
@rajat1994 
My immediate above comment is regarding that error only[...check it]https://github.com/zenodo/zenodo/pull/1043#issuecomment-292024471)

lnielsen:
@krzysztof @slint Was this problem fixed via the DOI versioning feature? If so, please close.

krzysztof:
It wasn't the problem still persists and this PR addresses it, but unfortunately theere is an issue with unit tests since the DOI check kicks in before the 403 (permission). I am not sure why.

PR #1045, opened 2017-04-02
global: change localhost to ZENODO_DOCKER_HOST in default configuration.

xiaom:
The purpose is to make it easier to setup development environment on non-linux systems (e.g., MacOS and Windows).
In these systems, docker runs inside a VM, and the host address is different from localhost.

### DIFF
--- a/zenodo/config.py
+++ b/zenodo/config.py
@@ -71,6 +71,8 @@ def _(x):
     """Identity function for string extraction."""
     return x
 
+# : Default docker host address
+ZENODO_DOCKER_HOST = os.environ.get("ZENODO_DOCKER_HOST", "localhost")
 #: Email address for support.
 SUPPORT_EMAIL = "info@zenodo.org"
 MAIL_SUPPRESS_SEND = True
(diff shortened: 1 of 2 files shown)


krzysztof:
Hey, thanks for the patch. As far as I know this problem does not exist in new OS X systems, as docker is binding to localhost ports.

To fix this for your version, you can redirect localhost ports to docker ports like this (keep this session alive):

``ssh -L 6379:localhost:6379 -L 5432:localhost:5432 -L 9200:localhost:9200 -L 5672:localhost:5672 docker@192.168.99.100``


With the proposed solution, the code/config will assume that we use docker - on production systems it will certainly not be true. I know that this would still work, since it falls back to "localhost" if the env variable is not set, but if it's not necessary it's better to have less assumptions about how we run the infrastructure.

krzysztof:
@xiaom Can you, instead of this patch, add a relevant "Note" to the installation guide for OS X users?

xiaom:
@krzysztof Thanks for the suggestion. I send a pull request https://github.com/zenodo/zenodo/pull/1048 for the installation guide.

zenodo/config.py
```diff
@@ -840,6 +842,10 @@ def _(x):
 
 # Search
 # ======
+#: Default ElasticSearch host
+SEARCH_ELASTIC_HOSTS = [
+    {"host": ZENODO_DOCKER_HOST}
+]
```
xiaom: Also, I think it would be helpful to explicitly configure ``SEARCH_ELASTIC_HOSTS`` here.
```
SEARCH_ELASTIC_HOSTS = [
    {"host": "localhost"}
]
```

krzysztof: Not needed since it's localhost by default:

https://github.com/inveniosoftware/invenio-search/blob/master/invenio_search/config.py#L38 (None from default Invenio config)
https://github.com/inveniosoftware/invenio-search/blob/master/invenio_search/ext.py#L107 (instantiated with None)
http://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch

krzysztof:
Closing since it has been resolved in #1048 
Thanks a lot for your contribution!

PR #1047, opened 2017-04-03
Use ; instead of , to separate authors

jougs:
The authors below the *Cite as* heading should be separated by a semicolon instead of a comma as the  authors right below the name of the publication are. The current display makes it hard to distinguish between first name, last name and the next author in the list.

zenodo-separator-bug

I dug through the code and believe that Zenodo's class [`ZenodoJSONSerializer`](https://github.com/zenodo/zenodo/blob/master/zenodo/modules/re/zenodo/modules/records/serializers/json.pycords/serializers/json.py#L37) should redefine its parent class' function [`_format_args()`](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/serializers/json.py#L42) to return a dictionary with `;` instead of `,` as separator.

This pull request fixes the issue.

### DIFF
--- a/zenodo/modules/records/serializers/json.py
+++ b/zenodo/modules/records/serializers/json.py
@@ -41,6 +41,28 @@ class ZenodoJSONSerializer(JSONSerializer):
     context to the request serializer.
     """
 
+    @staticmethod
+    def _format_args():
+        """Get JSON dump indentation and separates."""
+        # Ensure we can run outside a application/request context.
+        try:
+            pretty_format = \
+                current_app.config['JSONIFY_PRETTYPRINT_REGULAR'] and \
+                not request.is_xhr
+        except RuntimeError:
+            pretty_format = False
+
+        if pretty_format:
+            return dict(
+                indent=2,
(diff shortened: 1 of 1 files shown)


lnielsen:
@jougs Thanks for the PR. 

The place you changed is for the [JSON output](https://zenodo.org/record/44222/export/json) and would cause invalid JSON to be generated. Actually, the problem you are having is due to the default citation style which doesn't use semi-colon. We support some ~800 different citation styles, and each of them define whether to use semi-colon or comma to separate the authors. I.e. not something we can easily change.

jougs:
Thanks for reviewing and commenting. I now understand that this is much harder to fix in general and my fix was in the wrong place.

However, I still think that if the separator between authors is `,` it can't be `,` for separation of first and last name and that in those cases, at least first and last name should be displayed in reverse order and without the comma between the two components.

**current**
```
Last1, First1, Last2, First21 First22, Last3, First3, ...
```
**proposed**
```
First1 Last1, First21 First22 Last2, First3 Last3, ...
```

PR #1060, opened 2017-04-19
WIP global: fixed some of the failing unit tests

krzysztof:


docs/conf.py
```diff
@@ -38,6 +38,13 @@
 # Do not warn on external images.
 suppress_warnings = ['image.nonlocal_uri']
 
+# Ignore certain warn as error messages
+nitpick_ignore = [
```
krzysztof: We had to add this since the static fields in deposit API class are failing to resolve to a class-definition

krzysztof:
#1063 covers this already.

PR #1061, opened 2017-04-19
global: serializer and link generation fixes

slint:


zenodo/modules/records/serializers/schemas/common.py
```diff
@@ -384,109 +398,138 @@ class CommonRecordSchemaV1(Schema, StrictKeysMixin):
     """Common record schema."""
 
     id = fields.Integer(attribute='pid.pid_value', dump_only=True)
+    conceptid = SanitizedUnicode(
```
krzysztof: conceptrecid

PR #1065, opened 2017-04-25
GitHub Versioning Permissions

slint:


zenodo/modules/records/permissions.py
```diff
@@ -316,6 +343,17 @@ def has_update_permission(user, record):
     return has_admin_permission(user, record)
 
 
+def has_newversion_permission(user, record):
+    """Check if the user has permission to create a newversion for a record."""
+    recid = record.fetch_published()[0] if is_deposit(record) else record.pid
+    # GitHub records are treated differently in terms of versioning permissions
+    if is_github_versioned(recid):
+        return is_github_owner(user, recid)
```
krzysztof: it doesn't sync by default

zenodo/modules/records/templates/zenodo_records/box/edit.html
```diff
@@ -21,23 +21,42 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 -#}
+
+{% set has_doi = record.doi and record._deposit.id %}
+{% set edit_disabled = not has_doi %}
```
krzysztof: merge two variables into one?

Issue #856, opened 2016-10-14
records-ui: long filenames should be wrapped or truncated


krzysztof:
Long filenames are causing buttons to extend beyond the files div:
https://zenodo.org/record/154486
Perhaps we should truncate long filenames, e.g. "Long_long_long_filename.zip" -> "Long_long_long_file..." or break it mid filename ?

lnielsen:
I would go for wrap

ranjanam:
@krzysztof I would like to take up this issue.

rajat1994:
@krzysztof ... i would like to work on this issue if ranjanam is not working on it.

krzysztof:
@rajat1994 I guess you can start taking a look
@ranjanam are you actively working on this?

ranjanam:
@krzysztof Yes. I am working on this.

krzysztof:
@ranjanam I guess word wrap would be better.

jrios6:
Hi, is this bug resolved yet? @ranjanam @krzysztof

kishansag:
@krzysztof Hey, I have tried to fix this. Are you expecting something like this https://github.com/zenodo/zenodo/pull/1076?

krzysztof:
Fix has been integrated here https://github.com/zenodo/zenodo/pull/1241

PR #1076, opened 2017-05-02
Fix for records-ui: long filenames should be wrapped or truncated #856

kishansag:
Fix for records-ui: long filenames should be wrapped or truncated #856: Added css for link-text to wrap the text


krzysztof:
Closing since this PR has been integrated as part of big PR cleanup here: https://github.com/zenodo/zenodo/pull/1241

Thanks a lot for the contribution and sorry for the delay :)

PR #1084, opened 2017-05-08
deposit: disallow custom DOI for versioned records

slint:


slint:
@krzysztof cherry-pick or rebase at some point

PR #1099, opened 2017-05-23
installation: files-rest xss fix version bump

krzysztof:


setup.py
```diff
@@ -107,7 +107,7 @@
     'invenio-config>=1.0.0b1',
     'invenio-csl-rest>=1.0.0a1',
     'invenio-deposit>=1.0.0a7',
-    'invenio-files-rest>=1.0.0a14',
```
krzysztof: @lnielsen Why didn't we specify .post1 in setup.py requirements before?

PR #1111, opened 2017-05-25
frontpage: message display

lnielsen:


krzysztof:
I think you need to rebase

lnielsen:
I was related to using master on zenodo-migrator instead of a specific commit

PR #1143, opened 2017-05-31
WIP openaire: helper functions

lnielsen:
* Adds helper functions for determining OpenAIRE type of a record.

* Adds function to compute an OpenAIRE identifier as well as portal
  link.

Configuration var names needs to be coordinated with @slint.

### DIFF
--- a/zenodo/modules/openaire/helpers.py
+++ b/zenodo/modules/openaire/helpers.py
@@ -1,0 +1,118 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Zenodo.
+# Copyright (C) 2017 CERN.
+#
+# Zenodo is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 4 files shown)


zenodo/modules/openaire/helpers.py
```diff
@@ -0,0 +1,118 @@
...
+"""OpenAIRE related helpers."""
+
+from __future__ import absolute_import, print_function
+
+import hashlib
+
+from flask import current_app
+
+#: ID prefix for OpenAIRE publications.
+OPENAIRE_ID_PREFIX_PUBLICATION = 'od______2659'
```
lnielsen: these guys should not be here

slint: There's two solutions here: either moving these to `zenodo.config`, or adding `zenodo/modules/openaire/{config,ext}.py` and moving [this section](https://github.com/zenodo/zenodo/blob/master/zenodo/config.py#L272-L277) there along with the new stuff...

PR #1179, opened 2017-06-12
version: bump invenio-i18n to v1.0.0b3

xiaom:
I get the following warning when running `zenodo run  --reload --with-threads`.

```
DEBUG in babel [/Users/xiaom/anaconda/envs/zenodo/lib/python2.7/site-packages/invenio_i18n/babel.py:108]:
Compiled translations seems to be missing in
/Users/xiaom/anaconda/envs/zenodo/lib/python2.7/site-packages/invenio_access/translations.
```

Upgrading invenio_i18n resolves the issue.


krzysztof:
@xiaom Can you also bump it here? https://github.com/zenodo/zenodo/blob/master/setup.py#L119

We usually keep both synced, so that the lowest requrements on setup.py are the same as exact in requirements.txt

xiaom:
@krzysztof done.

PR #1199, opened 2017-06-17
deposit-ui: mockup button for "extraction" of a metadata

xiaom:
For now, the title can be updated as the filename after clicking the extracting button.

### DIFF
--- a/zenodo/modules/deposit/static/templates/zenodo_deposit/list.html
+++ b/zenodo/modules/deposit/static/templates/zenodo_deposit/list.html
@@ -48,6 +48,7 @@
       </th>
       <th>Size</th>
       <th class="text-center">Progress</th>
+      <th class="text-center">Extract</th>
       <th class="text-center">Delete</th>
     </tr>
     <tr class="sel-file" ng-repeat="f in filesVM.files | filter:fileSearch | orderBy:'key'">
@@ -63,13 +64,21 @@
         <span ng-show="f.errored" class="text-danger">Error</span>
         <span ng-show="f.completed" class="text-success"><i class="fa fa-check"></i></span>
       </td>
-      <td ng-show="f" class="text-center"><a href="#" ng-click="filesVM.remove(f)"><i class="fa fa-trash-o"></a></td>
+      <td class="text-center">
+        <button
+          class="btn btn-xs btn-default form-button {{ form.fieldHtmlClass }}"
+          ng-click="extractMetadata(f)"
+          extract-button
+        >
+        <i class="fa fa-magic" aria-hidden="true"></a></td>
+      </td>
+      <td ng-show="f" class"text-center"><a href="#" ng-click="filesVM.remove(f)"><i class="fa fa-trash-o"></a></td>
     </tr>
     <tr ng-show="(filesVM.files | filter:fileSearch).length == 0">
-      <td colspan="4" class="text-center text-muted"> No results.</td>
+      <td colspan="5" class="text-center text-muted"> No results.</td>
     </tr>
     <tr class="warning">
-      <td colspan="4">
+      <td colspan="5">
         <small class="text-muted">Note: File addition, removal or modification are not allowed after you have published your upload. This is because a Digital Object Identifier (DOI) is registered with <a href="http://www.datacite.org"><i class="fa fa-external-link"></i> DataCite</a> for each upload.</small>
       </td>
     </tr>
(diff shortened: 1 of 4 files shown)


krzysztof:
Hey, looks cool :)

Can you change the PR base from zenodo:master to zenodo:metadata_extraction (new feature branch I made)

zenodo/modules/deposit/static/templates/zenodo_deposit/list.html
```diff
@@ -63,13 +64,21 @@
         <span ng-show="f.errored" class="text-danger">Error</span>
         <span ng-show="f.completed" class="text-success"><i class="fa fa-check"></i></span>
       </td>
-      <td ng-show="f" class="text-center"><a href="#" ng-click="filesVM.remove(f)"><i class="fa fa-trash-o"></a></td>
+      <td class="text-center">
```
krzysztof: You can try to add ``ng-show="f.completed"`` to this td element so that the button is only visible when the file has been uploaded to the server.

krzysztof:
There seems to be an issue with "extractmetadata" link updating to ``f.links``.

When you create a new deposit, and choose a file (without uploading yet), the file in ``filesVM.files`` has only "key", "uri" and "multipart" properties (expected), see:

<a href="http://i.imgur.com/s9X9Jaa.png">
  
</a>

Now when you click "Upload file", it fills in relevant info:

<a href="http://i.imgur.com/WHR8NUZ.png">
  
</a>

but the "extractmetadata" link is not in "links" yet. At this point the button does not work because it cannot find the link (see errors in browser developer console). It does work only after you refresh the page.

Not sure how to solve it - it might require some changes upstream in invenio-files-js.

Alternatively, if possible see if you can use a completely separate endpoint which you can hardcode for now.

zenodo/modules/deposit/views.py
```diff
@@ -259,6 +260,18 @@ def current_datetime():
         'current_time': now.time(),
     }
 
+@blueprint.route(
+    '/record/extractmetadata/<key>/<version_id>',
+    methods=['POST']
+)
+def extractmetadata(key=None, version_id=None):
```
krzysztof: key is not necessary if you have ``version_id`` (one of PKs of ObjectVersion model). If we need the key later we can always get it: https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/models.py#L801

zenodo/modules/deposit/views.py
```diff
@@ -259,6 +260,18 @@ def current_datetime():
         'current_time': now.time(),
     }
 
+@blueprint.route(
+    '/record/extractmetadata/<key>/<version_id>',
```
krzysztof: Since this is already a POST-only endpoint perhaps we can pass the "version_id" in POST data?
i.e. this route would be only: '/record/extractmetadata/'.

zenodo/modules/deposit/views.py
```diff
@@ -259,6 +260,18 @@ def current_datetime():
...
+    methods=['POST']
+)
+def extractmetadata(key=None, version_id=None):
+    """Extract metadata for a given file
+
+    NOTE:
+    It should probably go to `invenio-files-rest` module or
```
krzysztof: I think the endpoint will go to separate module to keep it simple in the invenio-files-rest

xiaom:
@krzysztof I fix those issues.

zenodo/modules/deposit/templates/zenodo_deposit/edit.html
```diff
@@ -135,6 +135,45 @@
           };
         }]);
 
+        angular.module('invenioRecords.directives')
+        .directive('extractButton',['$rootScope', 'InvenioRecordsAPI', function($rootScope, InvenioRecordsAPI) {
+
+          function link($scope, elem, attrs, vm) {
+            $scope.extractMetadata = function(f) {
+                var url = `/record/extractmetadata/${ f.version_id }`
```
slint: This works for the prototyping (since we are all using modern browsers) but will need to be changed for production. Unfortunately until we improve our JS build process through Webpack, we will have to be checking [Can I Use...](https://caniuse.com/#feat=template-literals) before being adventurous :wink:

xiaom: Sure. I just added a TODO comment there.

https://github.com/xiaom/zenodo/blob/metadata/zenodo/modules/deposit/templates/zenodo_deposit/edit.html#L144-L146

PR #1208, opened 2017-06-22
metadata-extractor: initial grobid integration

xiaom:
I use this pdf for testing (https://zenodo.org/record/815512#.WUt4VIpJlTY)

It should fill the title, authors, description (using abstract) and keywords.
It works, but the code need to be polished.


zenodo/modules/deposit/extractor.py
```diff
@@ -25,12 +25,31 @@
...
 
 def zenodo_metadata_extractor(version_id):
-    metadata = {
-       "title": "<Title>",
-       "description": "version_id = {}".format(version_id),
-       "version_id": version_id
-    }
+    file_id = ObjectVersion.query.filter_by(version_id=version_id).one().file_id
+    file_instance = FileInstance.get(file_id)
+    with open(file_instance.uri) as pdf_file:
```
krzysztof: What python version are you using?
I had to add 'rb' parameter here (``open(file_instance.uri, 'rb')``) in order to work on Python 3.5, otherwise I got some utf-8 errors since by default it opened the PDF as text file.

xiaom: I use python 2.7.  I've fixed the problem following the practice here:

http://python-future.org/compatible_idioms.html#file-io-with-open

PR #1222, opened 2017-07-06
global: all dependencies upgrade

krzysztof:
* Upgrades all dependencies to the newest releases.

* Requires DB schema migration.


zenodo/config.py
```diff
@@ -172,6 +172,8 @@ def _(x):
 ACCOUNTS_SESSION_REDIS_URL = "redis://localhost:6379/2"
 #: Cache for storing access restrictions
 ACCESS_CACHE = 'zenodo.modules.cache:current_cache'
+#: Disable JSON Web Tokens
+ACCOUNTS_JWT_ENABLE=False
```
krzysztof: @lnielsen Do we enable JWT?

lnielsen: We need to test it properly first, since I'm not sure we have deposit-js released yet.

requirements.txt
```diff
@@ -26,211 +26,227 @@
...
+dulwich==0.17.3
+elasticsearch==2.4.1
 elasticsearch-dsl==2.2.0
-entrypoints==0.2.2
+entrypoints==0.2.3
+enum34==1.1.6
 execnet==1.4.1
-Flask==0.11.1
-Flask-Admin==1.4.2
+Flask==0.12.2
```
lnielsen: We should verify if the removal of mime type guessing has any impact on invenio-files-rest (don't think so, but better check)

requirements.txt
```diff
@@ -26,211 +26,227 @@
...
+jsmin==2.2.2
+jsonpatch==1.16
 jsonpointer==1.10
 jsonref==0.1
 jsonresolver==0.2.1
-jsonschema==2.5.1
-jupyter-core==4.2.0
-kombu==3.0.37
-Konch==2.2.0
-limits==1.2.0
```
lnielsen: Is limits not needed by Flask-Limiter?

requirements.txt
```diff
@@ -26,211 +26,227 @@
...
 mock==2.0.0
 msgpack-python==0.4.8
-nbconvert==4.2.0
-nbformat==4.1.0
+nbconvert==5.2.1
+nbformat==4.3.0
 ndg-httpsclient==0.4.2
 node-semver==0.1.1
-oauthlib==1.1.2
+oauthlib==2.0.2
```
lnielsen: I think there was some issues with one of the latest oauthlib releases being incompatible with Flask-Oauthlib. Not sure if this is still the case.

krzysztof: We are not using flask-limiter (invenio-app has it as a dependency, but we don't use it either).

requirements.txt
krzysztof: We are not using flask-limiter (invenio-app has it as a dependency, but we don't use it either).

requirements.txt
```diff
@@ -26,211 +26,226 @@
...
 Flask-Menu==0.5.1
-Flask-OAuthlib==0.9.3
+Flask-OAuthlib==0.9.4
 Flask-Principal==0.4.0
-Flask-Security-Fork==1.8.1
-Flask-SQLAlchemy==2.1
-Flask-WTF==0.13.1
+Flask-Security-Fork==2.0.1
+Flask-SQLAlchemy==2.2
+Flask-WTF==0.14.2
```
egabancho: Check this one before deploying because it has caused us some [headaches](https://github.com/inveniosoftware/invenio-oauthclient/issues/115) with Invenio-OAuthClient and the CERN app, I don't know if this is also the case for GitHub.

requirements.txt
lnielsen: We are using it, but it was recently moved to Invenio-App:

```
$ curl -i https://zenodo.org/api/records
HTTP/1.1 200 OK
Server: nginx/1.10.2
...
X-RateLimit-Limit: 5000
X-RateLimit-Remaining: 4999
X-RateLimit-Reset: 1499934808
Retry-After: 3599
...
```

slint:
Currently failing because of https://github.com/inveniosoftware/flask-celeryext/issues/29

PR #1223, opened 2017-07-06
metadata-extractor: fix mistakes in previous commit.

xiaom:


zenodo/modules/deposit/templates/zenodo_deposit/edit.html
```diff
@@ -145,31 +145,32 @@
               // TODO: we may need to avoid ES6 Template Literal in production to support
               // legacy browsers.
               // see https://caniuse.com/#feat=template-literals
+              var url = `/fileprocessor/${ filetype }/${ f.version_id }`
```
krzysztof: should be ``/filesprocessor/pdfgrobid/${ f.version_id}``

PR #1242, opened 2017-07-27
Footer update

krzysztof:


krzysztof:
Shortened "General Policies" to "Policies" since it was causing errors during page resizing.
It's probably fixable but requires more changes to the footer layout.

krzysztof:
Integrated in https://github.com/zenodo/zenodo/pull/1239

PR #1248, opened 2017-08-03
deposit-ui: fixed 404 error for new drafted deposits

krzysztof:


krzysztof:
This fixes the issue with deposit pages returning 404 for new deposit drafts, which might not have the DOI saved yet.

PR #1253, opened 2017-08-07
deposit: add alternate identifier to schema

slint:


krzysztof:
Indeed it's not possible to specify the alternate identifier from the deposit, but wouldn't it make the value end up in related_identifiers, instead of alternate_identifiers: https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/jsonschemas/records/record-v1.0.0.json#L399

PR #1254, opened 2017-08-09
metadata-extractor: extractor metadata modal

xiaom:
There are some unresolved UI problems.

1. Disable the apply button before the extraction is completed. `ng-disabled` does not work as I expected.
2. Close the modal when I apply the changes.  I use `ng-click="applyMetadata();$hide()" ` in the code. But it seems `$hide()` is not executed.

### DIFF
--- a/zenodo/modules/deposit/static/templates/zenodo_deposit/list.html
+++ b/zenodo/modules/deposit/static/templates/zenodo_deposit/list.html
@@ -80,15 +80,34 @@
           <div class="modal-dialog">
             <div class="modal-content">
               <div class="modal-header">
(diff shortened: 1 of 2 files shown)


krzysztof:
I've pushed some minor fixes. Overall looks nice.

I encountered this error: https://github.com/inveniosoftware/invenio-files-processor/issues/7
I can send you the file which causes this later.

1. fixed - I am disabling each individual button.
2. fixed - I don't close the modal at all since now there's multiple buttons. the user is supposed to close it with a "Close" button. But strangely - even before the changes the modal was closing for me ...

Issue #1238, opened 2017-07-26
null values from extractor


krzysztof:
Upload a PDF that is not a paper - extracted results is something like: 
```
{
   'title' null,
   'description': null,
  (...)
}
```
We should not attempt to replace the values in the form for those, as well as empty strings (not sure if that happens though).

krzysztof:
This should perhaps be tackled after this one: https://github.com/zenodo/zenodo/issues/1201
since in case of null values we would simply not show those on the modal form.

PR #1256, opened 2017-08-10
deposit-ui: hide a field if the extracted data is null and allow add …

xiaom:
…a single creator/keyword

Signed-off-by: Xiao Meng <xiaom.cs@gmail.com>

It should also resolve https://github.com/zenodo/zenodo/issues/1238

The UI still needs some polish.

First, the field "creators" and "keywords" are ordered array.  But we do not forbid users to add them in an arbitrary order.

Second, we want to disable the button "add" after users clicked it. To track whether an entry is already added, we may use a dictionary to track the state. Not sure whether this is a good solution.

### DIFF
--- a/zenodo/modules/deposit/static/templates/zenodo_deposit/list.html
+++ b/zenodo/modules/deposit/static/templates/zenodo_deposit/list.html
@@ -74,7 +74,6 @@
           data-target="#metadataModel"
           extract-button
         >
-       <!--"-->
         <i ng-show="f.completed" class="fa fa-magic" aria-hidden="true"></a></td>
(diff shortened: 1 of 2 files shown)


krzysztof:
Hey, I tested it, it works a bit better but I agree that there are still problems that need to be fixed.

> First, the field "creators" and "keywords" are ordered array. But we do not forbid users to add them in an arbitrary order.

I guess that's OK since it's up to user in the end.

> Second, we want to disable the button "add" after users clicked it. To track whether an entry is already added, we may use a dictionary to track the state. Not sure whether this is a good solution.

Or simply disable the button *always* after it has been clicked.

I will merge this since it improves the current state but we will need to get back to it again.

PR #1259, opened 2017-08-15
sipstore: bagit archiving of SIPs

krzysztof:


krzysztof:
Depends on https://github.com/inveniosoftware/invenio-sipstore/pull/33

zenodo/modules/sipstore/utils.py
```diff
@@ -0,0 +1,71 @@
...
+    :param sip: SIP which is to be archived
+    :type SIP: invenio_sipstore.models.SIP
+    :return: list of str
+    """
+    jsonmeta = SIPMetadata.query.get(
+        (sip.id, SIPMetadataType.get_from_name('json').id))
+    if jsonmeta is not None:
+        data = json.loads(jsonmeta.content)
+        revision = str(data['_deposit']['pid']['revision_id'])
+    else:
```
krzysztof: this means that the ``marcxml`` metadata content will be written under as if it was revision 0 (like for new records now), even though it's only the legacy records which have that. This is technically correct and safe since the migrated records's revision 0 is marcxml dump from legacy.

zenodo/config.py
```diff
@@ -254,6 +254,29 @@ def _(x):
...
+#: Set the SIPMetadata name formatter: "record-{json/marcxml}.{json/xml}"
+SIPSTORE_BAGIT_TAGS = [
+    ('Source-Organization', 'European Organization for Nuclear Research'),
+    ('Organization-Address', 'CERN, CH-1211 Geneva 23, Switzerland'),
+    ('Bagging-Date', None),  # Autogenerated
+    ('Payload-Oxum', None),  # Autogenerated
+    ('External-Identifier', None),  # Autogenerated
+    ('External-Description', ("BagIt archive of Zenodo record. "
+        "Description of the payload structure and data interpretation "
+        "available at DOI:10.5281/zenodo.841781")),
```
lnielsen: ``DOI:10.5281/zenodo.841781`` -> ``https://doi.org/10.5281/zenodo.841781``

zenodo/modules/sipstore/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+
+from celery import shared_task
+from invenio_db import db
+from invenio_sipstore.api import SIP as SIPApi
+from invenio_sipstore.archivers import BagItArchiver
+from invenio_sipstore.models import SIP
+
+
+@shared_task(ignore_result=True, rate_limit='100/m')
+def archive_sip(sip_uuid):
```
lnielsen: Shouldn't this task go in SIPStore?

zenodo/modules/sipstore/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+from __future__ import absolute_import
+
+from celery import shared_task
+from invenio_db import db
+from invenio_sipstore.api import SIP as SIPApi
+from invenio_sipstore.archivers import BagItArchiver
+from invenio_sipstore.models import SIP
+
+
+@shared_task(ignore_result=True, rate_limit='100/m')
```
lnielsen: Necessary to rate limit as long as we push it to the low priority queue?

krzysztof: Perhaps not, but then this task might then throttle the file checks tasks if we have a rate of >100 publishings a minute and if the total archiving task time takes longer than a minute. The low priority queue does not have any rate limiting mechanism by itself. The file check tasks are 'rate-limited' by a separate scheduler task, and here I don't have one.

zenodo/modules/sipstore/tasks.py
krzysztof: I can, but the task signature (rate limit, retries etc.) is specific to our configuration and assumes archiving by bagit. Not sure if it's reusable, but at least the body of the task could be moved there.

lnielsen: But do we need to rate limit it now that we have a low priority queue?

zenodo/modules/sipstore/tasks.py
krzysztof: Nonetheless, I have removed rate_limiting for the time migration though, we can put it back later.

PR #1263, opened 2017-08-16
WIP: Packages upgrades + Invenio-App/Cache/XRootD

lnielsen:
Things to verify:
- [x] Date granularity harvesting from OAI-PMH (marshmallow/oaiserver upgraded)
- [x] OAIServer updated timestamp.
- [x] Access control in general
- [x] Security Headers: Disable Content Security Policy (and create issue to enable it).
- [x] Security Headers: in general check that the security headers don't mess up the site.
- [x] Check that there's no more Celery 4 config variable warnings.
- [x] OAuth2Server: new scope user:email
- [x] OAuth2Server: example link to authorise end point
- [x] OAuth2Server/Search: badly encoded URL strings (error message)
- [x] OAuthClient: should not redirect to linked accounts anymore when logging from frontage.
- [x] Frontpage caching
- [x] JSON -> JSONB alembic upgrade
- [x] JSONSchemas UI/API endpoint (config variables).
- [x] Search (due to bouncing search results)

For deployment:
- [ ] Rename `/opt/zenodo/var/instance/zenodo.cfg` to `invenio.cfg`
- [ ] Removal of some old dependencies.
- [ ] Alembic upgrades

Depends on https://github.com/inveniosoftware/invenio-base/pull/121


requirements.txt
```diff
@@ -29,127 +29,123 @@
 -e git+https://github.com/zenodo/zenodo-migrator.git#egg=zenodo-migrator
```
lnielsen: When we deploy we have to take care of removing old dependencies. In particular Flask-Security-Fork and many test dependencies.

requirements.pinned.txt
```diff
@@ -24,3 +24,5 @@
 
 # 2.13.5 break Invenio-OAIServer date granularity validation.
 marshmallow==2.13.4
+# 0.14 have changed CSRF validation.
+Flask-WTF==0.13.1
```
lnielsen: Issues with 0.14 should have been fixed in upstream Invenio, but not full sure so I leave it pinned.

requirements.pinned.txt
```diff
@@ -22,5 +22,5 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
-# 2.13.5 break Invenio-OAIServer date granularity validation.
-marshmallow==2.13.4
+# 0.14 have changed CSRF validation.
```
lnielsen: Issues with 0.14 should have been fixed in upstream Invenio, but not full sure so I leave it pinned.

lnielsen:
List of packages that should be uninstalled from qa/prod:

```
alabaster
apipkg
check-manifest
coverage
docutils
Flask-Cache
flask-konch
Flask-Security-Fork
funcsigs
imagesize
isort
konch
mock
pbr
pep8
py
pydocstyle
pytest
pytest-cache
pytest-cov
pytest-flask
pytest-mock
pytest-pep8
pytest-random-order
selenium
snowballstemmer
Sphinx
```

slint:
depends on inveniosoftware/invenio-communities/pull/110

lnielsen:
Merged inveniosoftware/invenio-communities#110

zenodo/modules/theme/templates/zenodo_theme/security/login_user.html
```diff
@@ -26,8 +26,8 @@
 
 {%- block form_outer %}
   <div class="social-signup">
-    <a href="{{url_for('invenio_oauthclient.login', remote_app='github')}}" class="btn btn-default btn-lg btn-block"><i class="fa fa-github"></i> {{_('Log in with GitHub')}}</a>
-    <a href="{{url_for('invenio_oauthclient.login', remote_app='orcid')}}" class="btn btn-default btn-lg btn-block"><img src="{{ url_for('static', filename='img/orcid.svg')}}" height="18px"> {{_('Log in with ORCID')}}</a>
+    <a href="{{url_for('invenio_oauthclient.login', remote_app='github', next=(request.args.get('next') or request.referrer))}}" class="btn btn-default btn-lg btn-block"><i class="fa fa-github"></i> {{_('Log in with GitHub')}}</a>
```
lnielsen: I think we will need to sanitize the next URL to include only local URLs. We have a utility function somewhere to do this, just don't remember where right now (accounts, oauthclient or oauth2server I believe).

slint: I assume [invenio_oauthclient.utils.get_safe_redirect_target](https://github.com/inveniosoftware/invenio-oauthclient/blob/2359e19caa9ef1063a4b16f0d1dc648755c5823a/invenio_oauthclient/utils.py#L175-L184), which is already used inside the [`/oauth/login/<remote_app>`](https://github.com/inveniosoftware/invenio-oauthclient/blob/7ef6bfbcf9a4eec3ee78f6fb46edfd7495d82fb8/invenio_oauthclient/views/client.py#L89) view. Also this is kind of already protected because it's passed to the OAuth workflow which will reject it in case it's not a [sub-path of the default `redirect_uri`](https://developer.github.com/apps/building-integrations/setting-up-and-registering-oauth-apps/about-authorization-options-for-oauth-apps/#redirect-urls)

lnielsen: Ah, correct! I was looking in ``authorized`` view :-( Yes, you're right it's already protected.

Regarding the GitHub callback URL that's a separate thing. The ``next`` parameter is something completely isolated to Invenio. I.e. GitHub redirects to /oauth/authorized/github/, which redirects to the next URL.

PR #1264, opened 2017-08-16
Separate Javascript bundle for Zenodo deposit form

slint:


slint:
As far as I could test things out, deposit form works, though it might be worth keeping in a separate branch before merging into master

lnielsen:
👍 Perhaps let's pool it with files extractor and researcher profiles deployment so we can do some thorough testing of it.

slint:
These changes were merged as part of https://github.com/zenodo/zenodo/pull/1365

PR #1277, opened 2017-08-25
openaire: addition of a fixed funder list

krzysztof:
* Adds a fixutre containing only the supported funders.


krzysztof:
@slint openaire_sprint is the feature branch.

PR #1284, opened 2017-09-01
global: files jsonschema fixes

krzysztof:
* Adds 'file_id' as files schema key and introduces requried keys.

* Refactoring of jsonschema compilers and file's 'source' schemas.


krzysztof:
Main change is addition of "file_id" and required keys. It's no longer necessary to compile files jsonschema as the file-v1.0.0 could have been described explicitly in the first place (there was no sharing of common file_src code).

slint:
Apparently we're making an actual request in tests to the OpenAIRE Indexing API (fortunately, at the testing endpoint). It's probably caused by some publish signal firing the task.

zenodo/modules/records/jsonschemas/records/file-v1.0.0.json
```diff
@@ -3,51 +3,71 @@
...
           },
           "version_id": {
+            "description": "UUID of the ObjectVersion object.",
             "type": "string"
           }
         },
+        "required": [
+          "bucket",
+          "checksum",
+          "file_id",
```
slint: Could this lead on issues (eg. merge conflicts) with re-publishing old records that are missing some of the fields?

krzysztof: I don't think so. In general merging of old records takes the record's metadata at the revision specified by the deposit (_deposit.pid_revision_id), computes a diff and applies the diff on the newest record metadata. Conflicts occur if the newest record's metadata conflicts with the diff of two jsons, but in this case I don't think it's possible, since somehow user would have to edit the deposits "_files"

krzysztof:
Updated support form text:
<a href="https://i.imgur.com/xbzhDsT.png">
  
</a>

PR #1298, opened 2017-09-26
docs: installation guide with NVM

krzysztof:

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -167,21 +167,54 @@ Media assets
 
 Next, we need to build the assets for the Zenodo application.
 
-To compile Zenodo assets you will need some NPM pacakges and asset building
-tools: NodeJS, SASS, CleanCSS, UglifyJS and RequireJS.
-The easiest way is to install them system-wide, and in the specific versions we
-have pinned. You can do that by executing:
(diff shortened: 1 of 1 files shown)


krzysztof:
see here https://github.com/krzysztof/zenodo/blob/installation_fixes/INSTALL.rst

INSTALL.rst
```diff
@@ -167,21 +167,54 @@ Media assets
...
+can install the asset tools system-wide (with ``sudo``), by executing:
 
 .. code-block:: console
 
    (zenodo)$ sudo ./scripts/setup-npm.sh
 
 Take a look in the script above to see which commands are being run.
-Since those pacakges are installed with ``-g`` flag (system-wide),
-you will need to run the command above with ``sudo``.
+Use of ``sudo`` is required because of the ``-g`` flag for global installation.
```
slint: There's also this solution for avoiding `sudo` and installing npm in the user's home directory: https://github.com/sindresorhus/guides/blob/master/npm-global-without-sudo.md

krzysztof: I guess to not further complicate this small step in a guide I will leave it as is, since installing npm in home directory is almost the same efforts as using nvm at this point.

PR #1301, opened 2017-09-28
Deposit version field

krzysztof:


krzysztof:
This introduces the version field back. to be contined once we fix the UI and migrate github stuff

Issue #1308, opened 2017-10-04
deposit-ui: ORCID for contributors


slint:

krzysztof:
This was already integrated into master.

PR #1309, opened 2017-10-04
ORCID field for contributors + supervisors, and minor fixes

slint:


slint:
Closes #1308

zenodo/modules/spam/templates/zenodo_spam/delete.html
```diff
@@ -104,7 +104,7 @@
         <div class="panel-body">
           User has <strong>{{ rec_count }}</strong> records indexed in elasticsearch.
 
-          <a href="{{ url_for('invenio_search_ui.search', q='owners: ' + (user.id|string)) }}"><strong>Full record search result</strong></a>
+          <a href="{{ url_for('invenio_search_ui.search', q=('owners:' + (user.id|string))) }}"><strong>Full record search result</strong></a>
```
krzysztof: curious - what was the bug here?

slint: I think the space didn't allow the user ID to be parsable (ie as `owners:12345`, but rather just a generic argument (ie. `owners:` and `12345`), so results weren't consistent.

Issue #1333, opened 2017-11-03
records: language missing from marcxml


lnielsen:
E.g. 041__a:eng

krzysztof:
https://github.com/zenodo/zenodo/commit/080fc3aee5103bf5d0904685b74b1bea7954e946#diff-c01412a79ee78db1d35383c103f5ef65

PR #1335, opened 2017-11-06
records: marcxml language serialization

krzysztof:

### DIFF
--- a/zenodo/modules/records/serializers/to_marc21/rules.py
+++ b/zenodo/modules/records/serializers/to_marc21/rules.py
@@ -99,7 +99,7 @@ def reverse_oai(dummy_self, key, value):
 
 @to_marc21.over('856', '^conference_url$')
 @utils.filter_values
-def reverse_conkerence_url(dummy_self, key, value):
+def reverse_conference_url(dummy_self, key, value):
     """Reverse - Meeting."""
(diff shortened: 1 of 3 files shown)


krzysztof:
Closes https://github.com/zenodo/zenodo/issues/1333

zenodo/modules/records/serializers/to_marc21/rules.py
```diff
@@ -123,7 +123,12 @@ def reverse_files(dummy_self, key, value):
     }
 
 
-# @to_marc21.over('foobar', '^foobar$')
-# def to_leader_field(dummy_self, key, value):
-#     """Reverse - Files."""
-#     return value
+@to_marc21.over('041', '^language$')
```
slint: There seems to also be a `008/35-37` field [mentioned here](https://www.loc.gov/marc/bibliographic/bd041.html) and [here](https://www.loc.gov/marc/bibliographic/bd008a.html) (search for `008/35-37`) which is about language. I'm no MARC21 expert though...

krzysztof: That is complementary, but it looks like we don't serialize the field 008 in the first place. If we did, we could put the the language code into it.

Issue #1205, opened 2017-06-19
records: meta-tags problem for zotero


lnielsen:
E.g. on https://doi.org/10.5281/zenodo.45756 

this 

``<meta name="citation_author" content="Tempel" /> ``

should be

``<meta name="citation_author" content="Tempel, Wolfram" /> ``

krzysztof:
Fix merged to master branch

PR #1336, opened 2017-11-06
record-ui: HTML Highwire meta author name

krzysztof:
* Fixes the serialization of author names in Highwire
  meta tags.


krzysztof:
Closes https://github.com/zenodo/zenodo/issues/1205

It seems that "name" is the default field for our metadata, and "familyname" "givenname" is supported but doesn't exist in newer records. It's probably only for legacy records like this one (see the Admin-view of the DB JSON): https://zenodo.org/record/45756#.WgB9zHXysUH

Nonetheless, it seems that even for legacy records we're migrated it to "name" field too.

PR #1339, opened 2017-11-09
deposit-ui: opening links on deposit page in new tab

krzysztof:


krzysztof:
Reported by user. Clicking any link on the deposit page will open it in the new tab (saves user's from losing their changes).

slint:
Apparently, there are some security issues with doing this, and adding `rel="noopener noreferrer"` solves this... [Source](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/), courtesy of HN...

PR #1346, opened 2017-11-15
Openaire new funders

krzysztof:


krzysztof:
Depends on + release: https://github.com/inveniosoftware/invenio-openaire/compare/master...krzysztof:new_funders?expand=1

PR #1369, opened 2017-12-12
Secure & update hyperlinks

katrinleinweber:
Hello!

This updates some links the HTTPS version of sites that either forward automatically, or have it separately but working. Similarly, the old `dx.doi.org` is redirected automatically, but let's switch it to `https://doi.org` at the source.

About the `...API update`: In case the tests should cover the non-HTTPS links as well, please let me know and I'll revert those lines. If you on the other hand agree to only continue testing the HTTPS links, I would push another (few) commit(s). `d-nb.info` for example would be securely linked to at the source.

Cheers :-)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,7 @@ any size, any format and from any science.
 Powered by Invenio
 ------------------
 Zenodo is a small layer on top of
-`Invenio <http://github.com/inveniosoftware/invenio>`_, a ​free software
+`Invenio <https://github.com/inveniosoftware/invenio>`_, a ​free software
 suite enabling you to run your own ​digital library or document repository on
 the web.
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -150,8 +150,8 @@ html_theme_options = {
     'github_banner': True,
     'show_powered_by': False,
     'extra_nav_links': {
-        'zenodo@GitHub': 'http://github.com/zenodo/zenodo',
-        'zenodo@PyPI': 'http://pypi.python.org/pypi/zenodo/',
+        'zenodo@GitHub': 'https://github.com/zenodo/zenodo',
+        'zenodo@PyPI': 'https://pypi.python.org/pypi/zenodo/',
     }
 }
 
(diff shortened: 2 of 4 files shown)


tests/unit/records/test_schemas_datacite.py
```diff
@@ -97,7 +97,7 @@ def test_full(db, record_with_bucket, recid_pid):
                 "nameIdentifier": {
                     "nameIdentifier": "0000-0002-1825-0097",
                     "nameIdentifierScheme": "ORCID",
-                    "schemeURI": "http://orcid.org/"
+                    "schemeURI": "https://orcid.org/"
```
krzysztof: Looks like the orcid changes need some more fixes in serializer: https://travis-ci.org/zenodo/zenodo/jobs/315233527#L6482

zenodo/modules/fixtures/data/funders.json
```diff
@@ -11,7 +11,7 @@
     },
     "name": "National Health and Medical Research Council",
     "parent": {
-      "$ref": "http://dx.doi.org/10.13039/501100003921"
+      "$ref": "https://doi.org/10.13039/501100003921"
```
krzysztof: We resolve those DOIs "internally" with: https://github.com/inveniosoftware/invenio-openaire/blob/master/invenio_openaire/resolvers/funders.py#L44-L45

which should support both dx.doi.org and doi.org, but just in case this fails in a test somewhere, it might require some fixes upstream.

krzysztof:
> About the ...API update: In case the tests should cover the non-HTTPS links as well, please let me know and I'll revert those lines. If you on the other hand agree to only continue testing the HTTPS links, I would push another (few) commit(s). d-nb.info for example would be securely linked to at the source.

We can use HTTPS in test cases where possible, unless it's also trivial to test both (e.g. add few extra lines in test). Know that this might require some extra changes in serializers/code and sometimes even outside Zenodo (Invenio packages).

katrinleinweber:
Hi @krzysztof! I think the "extract"ion is complete. [That one error](https://travis-ci.org/zenodo/zenodo/jobs/328397986#L2846) is a mystery to me: `ERROR at setup of test_record_modified_while_edit`.

PR #1384, opened 2017-12-15
JSON-LD serializer

slint:


zenodo/modules/records/templates/zenodo_records/record_detail.html
```diff
@@ -73,7 +73,7 @@ <h1 itemprop="name">{{ record.title }}</h1>
         <h5>{{_('Thesis supervisor(s)')}}</h5>
         <p>{{ authors(record.thesis.supervisors) }}</p>
       {%- endif %}
-      <p itemprop="description">{{ record.description|safe }}</p>
+      <p>{{ record.description|safe }}</p>
```
krzysztof: @lniensen is it OK to pipe the JSON (string) from a serializer output through ``safe`` here?

zenodo/modules/records/templates/zenodo_records/record_detail.html
```diff
@@ -155,4 +155,8 @@ <h5>{{_('Thesis supervisor(s)')}}</h5>
 </script>
 <script type="text/javascript" src="//s7.addthis.com/js/300/addthis_widget.js#pubid=ra-5137aff66ad9c2a1"></script>
 <script type='text/javascript' src='https://d1bxh8uas1mnw7.cloudfront.net/assets/embed.js'></script>
+{% set jsonld_serialization = record | serialize_record(pid, 'schemaorg_jsonld_v1')%}
+<script type='application/ld+json'>
+{{ jsonld_serialization|safe }}
```
krzysztof: @lniensen is it OK to pipe the JSON (string) from a serializer output through ``safe`` here?

krzysztof:
@lnielsen we've removed most of the microdata from tags, I suppose whoever needed it before, can now parse the JSONLD at the bottom of the page within <script></script>?

zenodo/modules/records/serializers/__init__.py
```diff
@@ -87,6 +88,8 @@
 citeproc_v1 = CiteprocSerializer(csl_v1)
 #: OpenAIRE JSON serializer
 openaire_json_v1 = JSONSerializer(RecordSchemaOpenAIREJSON, replace_refs=True)
+#: JSON-LD serializer
+schemaorg_jsonld_v1 = JSONSerializer(CreativeWorkV1, replace_refs=True)
```
krzysztof: we're actually using regular serializer since there were problems with getting the correct serialized output from JSONLD one, and it was also making a request to compact/expand the json

zenodo/modules/records/serializers/schemaorg.py
```diff
@@ -0,0 +1,85 @@
...
+        'image-photo': schemas.Photograph,
+        'image-other': schemas.ImageObject,
+    }
+
+    def dump(self, obj, context=None):
+        """Serialize object with schema."""
+        data = obj['metadata']
+        obj_type = ObjectType.get_by_dict(data['resource_type'])
+        internal_id = obj_type['internal_id']
+        schema_cls = self.SCHEMA_ORG_TYPES.get(
```
krzysztof: maybe we could skip the double config (here and objecttypes.json) and resolve this with something like ``schemas.getattr(obj_type['schema.org'].split('https://schema.org/')[1])``?

slint: The only problem I see with that is that if somebody changes some `schema.org` key from `objecttypes.json`, the app will still be runnable. If we have tests for all `internal_id -> schema.org` types then it's not an issue of course.

zenodo/modules/records/serializers/schemaorg.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class ZenodoSchemaOrgSerializer(ZenodoJSONSerializer):
+    """Zenodo schema.org serializer.
+
+    Serializes the record using the appropriate marshmallow schema based on
+    its schema.org type.
+    """
+
+    def funcname(parameter_list):
```
krzysztof: is this required?

slint: Nope, just forgot it :see_no_evil:

zenodo/modules/records/serializers/schemaorg.py
lnielsen: 1) I would do the mapping based on ``obj_type['schema.org']`` instead of ``obj_type['internal_id']``. That should at least make the mapping somewhat smaller.
2) However, probably something like @krzysztof suggests is as good:
``schema_cls = getattr(schemas, obj_type['schema.org'].split('https://schema.org/')[1], schemas.CreativeWork)``

zenodo/modules/records/templates/zenodo_records/record_detail.html
lnielsen: I think it should be safe, since it will be a valid JSON encoding - i.e. no escaping necessary.

zenodo/modules/records/serializers/schemas/schemaorg.py
```diff
@@ -172,31 +181,38 @@ def get_url(self, obj):
...
+        ids = _serialize_identifiers(relids, {'hasPart'})
+        return ids or missing
 
     def get_sameAs(self, obj):
         """Get identical identifiers of the record."""
         relids = obj.get('metadata', {}).get('related_identifiers', [])
-        return [i['@id']
+        ids = [i['@id']
                 for i in _serialize_identifiers(relids, {'isIdenticalTo'})]
+        relids = obj.get('metadata', {}).get('alternate_identifiers', [])
```
krzysztof: adding all related identifiers as sameAs

PR #1402, opened 2018-01-11
global: file notifications and fixes

slint:


zenodo/modules/deposit/api.py
```diff
@@ -245,11 +245,12 @@ def _get_auto_added(self, record):
...
         """Snapshot bucket and add files in record during first publishing."""
         if self.files:
-            file_uuids = []
+            file_uuids = set()
             for f in self.files:
                 fs, path = f.file.storage()._get_fs()
-                if not fs.exists(path):
-                    file_uuids.append(str(f.file.id))
+                if not (fs.exists(path) and
+                        f.file.verify_checksum(throws=False)):
```
krzysztof: Shouldn't this condition be
```python
if not fs.exists(path) or not f.file.verify_checksum(throws=False):
```
but also assuming also that ``verify_checksum`` fails when file doesn't exist (not sure about that, needs checking), perhaps it should be simply:
```python
if not f.file.verify_checksum(throws=False):
```

zenodo/config.py
```diff
@@ -73,6 +73,12 @@ def _(x):
     """Identity function for string extraction."""
     return x
 
+
+#: System sender email address
+ZENODO_SYSTEM_SENDER_EMAIL = 'system@zenodo.org'
+#: Email address of admins
+ZENODO_ADMIN_EMAIL = 'zenodo-admin@cern.ch'
```
krzysztof: maybe keep the default names (besides license etc.), strictly zenodo-specific, but not cern-specific? e.g.: 'zenodo-admin@zenodo.org` and move the cern email to the instance config?

zenodo/modules/utils/tasks.py
```diff
@@ -263,3 +270,71 @@ def update_search_pattern_sets():
     """
     oaisets = OAISet.query.filter(OAISet.search_pattern.isnot(None))
     chain(make_oai_task_group(oais) for oais in oaisets).apply_async()
+
+
+def format_file_integrity_report(report):
+    """Format the email body for the file integrity report."""
```
krzysztof: Maybe use a jinja template as for other emails? Can be done later - just create an issue for this.

zenodo/modules/utils/tasks.py
```diff
@@ -263,3 +270,71 @@ def update_search_pattern_sets():
...
+        entry = {'file': f}
+        for o in f.objects:
+            entry['filename'] = o.key
+            # Find records/deposits for the files
+            rb = RecordsBuckets.query.filter(
+                RecordsBuckets.bucket_id == o.bucket_id).one_or_none()
+            if rb:
+                if is_deposit(rb.record.json):
+                    entry['deposit'] = rb.record.json
+                elif is_record(rb.record.json):
```
krzysztof: Something like ``else:  # must be a record if not the deposit`` ?

krzysztof:
All are minor comments, except the condition in the deposit which I'm not sure is correct

zenodo/modules/utils/tasks.py
krzysztof: Actually this might then include deleted files..

PR #1410, opened 2018-01-15
github: removed records bugfix

krzysztof:


krzysztof:
Depends on https://github.com/inveniosoftware/invenio-github/pull/58

Requires data migration

zenodo/modules/deposit/utils.py
```diff
@@ -191,6 +192,10 @@ def delete_record(record_uuid, reason, user):
     })
     record.commit()
 
+    # Mark the relevant GitHub Release as deleted
+    for ghr in record.model.github_releases:
+        ghr.status = ReleaseStatus.DELETED
```
slint: I'm not 100% sure but a `db.session.add(ghr)` might be needed here... If you tested and it works then :shipit:

PR #1429, opened 2018-03-01
requirements: update versions

ChiaraBi:


requirements.txt
```diff
@@ -105,7 +105,7 @@ invenio-assets==1.0.0b7
 invenio-base==1.0.0b1
 invenio-cache==1.0.0b1
 invenio-celery==1.0.0b3
-invenio-communities==1.0.0a18
+invenio-communities==1.0.0a19
```
slint: Since this version uses `marshmallow>=2.15.0`, `marshmallow` has to be bumped to this version here as well (and `setup.py`).

Another thing to try is also removing [the `--pre` flag from `.travis.yml`](https://github.com/zenodo/zenodo/blob/03dee2c29297da11964bea3fed53d0bb6a1a75dc/.travis.yml#L89), since it is causing beta versions of packages to be installed (ie. `marshmallow v3.0.0b7` in this case), and basically makes builds unpredictable.

PR #1433, opened 2018-03-06
Update datacitev4

ChiaraBi:


zenodo/modules/records/admin.py
```diff
@@ -0,0 +1,124 @@
...
+                    if from_date > until_date:
+                        flash("Error: the 'From' date should precede the 'Until' date.")
+                    else:
+                        pids_count = find_registered_doi_pids(from_date,
+                                                                    until_date,
+                                                                    current_app.config['ZENODO_LOCAL_DOI_PREFIXES']).count()
+                        task_details = dict(
+                            total_pids=pids_count
+                        )
+                        time = pids_count/DATACITE_UPDATING_RATE_PER_HOUR
```
slint: It's better to use `current_app.config['DATACITE_UPDATING_RATE_PER_HOUR']`, since this might be overriden by some config in e.g. `invenio.cfg`

zenodo/config.py
```diff
@@ -743,6 +752,8 @@ def _(x):
                 'zenodo.modules.records.serializers.bibtex_v1_response'),
             'application/x-datacite+xml': (
                 'zenodo.modules.records.serializers.datacite_v31_response'),
+            'application/x-datacite+xmlv41': (
```
slint: @lnielsen @krzysztof `application/x-datacite+xmlv41` or `application/x-datacite-v41+xml`

zenodo/modules/records/data/datacite-v4.1-full-example.json
```diff
@@ -0,0 +1,235 @@
+{
```
slint: Is this file (and other example files under this folder) being used by tests? If yes, it should be somewhere under the `tests` folder.

zenodo/modules/records/serializers/schemas/datacite.py
```diff
@@ -313,3 +319,134 @@ def inject_type(c):
                 })
 
         return items
+
+    def get_related_identifiers(self, obj):
+        """Resource type."""
+        items = super(DataCiteSchemaV1, self).get_related_identifiers(obj)
+        print('get_related_identifiers:')
```
slint: Probably used for debugging, just remember to remove before merging

zenodo/modules/records/tasks.py
```diff
@@ -46,3 +54,60 @@ def update_expired_embargos():
...
     indexer.process_bulk_queue()
+
+
+@shared_task(ignore_result=True, rate_limit='100/m')
+def update_datacite_metadata(pid_value, object_uuid, job_id):
+    """Update DataCite metadata of a single PersistentIdentifier."""
+
+    task_details = current_cache.get('update_datacite:task_details')
+
+    if task_details is None or job_id != task_details['job_id']:
```
slint: `job_id` is a string and `task_details['job_id']` a `uuid.UUID`.

```python
import uuid
u = uuid.uuid4()
assert str(u) != u
```

So, you can either store string version of it in Redis or cast to `str` here.

zenodo/modules/records/templates/zenodo_records/update_datacite.html
```diff
@@ -0,0 +1,86 @@
+{#
+# This file is part of Zenodo.
+# Copyright (C) 2016 CERN.
```
slint: Should be 2018. Take a look at other files as well that you either created or modified.

zenodo/modules/records/admin.py
```diff
@@ -0,0 +1,124 @@
...
+from wtforms.validators import DataRequired
+
+from zenodo.config import DATACITE_UPDATING_RATE_PER_HOUR
+from zenodo.modules.records.utils import find_registered_doi_pids
+
+
+class UpdateDataciteForm(FlaskForm):
+    """Form for updating datacite metadata."""
+
+    field_sets = ['from_date', 'until_date']
```
slint: These are not really needed. The reason they exist everywhere is because of [`invenio-communities`](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/forms.py#L40), which had a pretty large form and is dynamically rendered in Jinja.

For small forms it's okay to manually render fields like you've done

zenodo/modules/records/utils.py
```diff
@@ -98,3 +105,37 @@ def is_valid_openaire_type(resource_type, communities):
...
+    )
+
+    query.filter(or_(PersistentIdentifier.pid_value.like(prefix + '%') for prefix in prefixes))
+
+    query.order_by(PersistentIdentifier.updated)
+
+    return query
+
+
+def xsd41():
```
slint: It's nice to have this here, though we should consider at some point moving it to `inveniosoftware/datacite` utils. WDYT @lnielsen?

tests/unit/records/test_records_tasks.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # no task_details on Redis cache
+    schedule_update_datacite_metadata(1)
+    assert_datacite_calls_with_missing_data()
+    new_update_date1 = doi_pids[0].updated
+    assert update_date == new_update_date1
+
+    task_details = dict(
+        job_id=uuid.uuid4(),
+        from_date=datetime.strptime('Jan 1 2015  1:33PM', '%b %d %Y %I:%M%p'),
```
slint: It's better to define times explicitly via `datetime(2015, 1, 1, 13, 33)`

tests/unit/records/test_records_tasks.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # Create a reserved recid
+    recid1, r1 = create_versioned_record('352543', conceptrecid)
+
+    # no registered local DOIs
+    schedule_update_datacite_metadata(1)
+    assert_datacite_calls_with_missing_data()
+
+    doi_pids = PersistentIdentifier.query.filter(
+        PersistentIdentifier.pid_value.contains('10.5072/zenodo.352543'))
```
slint: This would be more explicit and safe:

```python
doi_pid = PersistentIdentifier.query.filter(PersistentIdentifier.pid_value == '10.5072/zenodo.352543').one()
```

zenodo/modules/records/admin.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class UpdateDataciteView(BaseView):
+    """View for updating datacite metadata."""
+
+    @expose('/', methods=('GET', 'POST'))
+    def data_cite(self):
+        """."""
+        form = UpdateDataciteForm()
+        details_form = FlaskForm()
```
slint: Maybe a better name like `cancel_or_new_task_form`

zenodo/modules/records/admin.py
```diff
@@ -0,0 +1,124 @@
...
+        description=_('Required.'),
+        validators=[DataRequired()],
+    )
+
+
+class UpdateDataciteView(BaseView):
+    """View for updating datacite metadata."""
+
+    @expose('/', methods=('GET', 'POST'))
+    def data_cite(self):
```
slint: `data_cite` -> `update_datacite` or something in these lines

zenodo/config.py
slint: I guess we go with `application/x-datacite-v41+xml`, which seems more proper

zenodo/modules/records/utils.py
slint: Let's remember to make an issue after merging this PR for moving it to `inveniosoftware/datacite`

zenodo/modules/records/serializers/__init__.py
```diff
@@ -105,6 +106,9 @@
 #: DataCite v3.1 record serializer for individual records.
 datacite_v31_response = record_responsify(
     datacite_v31, 'application/x-datacite+xml')
+#: DataCite v4.1 record serializer for individual records.
+datacite_v41_response = record_responsify(
+    datacite_v41, 'application/x-datacite+xml')
```
slint: Missed one :see_no_evil: (Should be `application/x-datacite-v41+xml`)

zenodo/config.py
```diff
@@ -219,6 +221,13 @@ def _(x):
         'task': 'zenodo.modules.utils.tasks.file_integrity_report',
         'schedule': crontab(minute=0, hour=7),  # Every day at 07:00 UTC
     },
+    'datacite-metadata-updater': {
+        'task': 'zenodo.modules.records.schedule_update_datacite_metadata',
+        'schedule': timedelta(hours=1),
+        'kwargs': {
+            'max_count': 1000,
```
slint: You can use `'max_count': DATACITE_UPDATING_RATE_PER_HOUR` here to be explicit about the value.

zenodo/modules/records/httpretty_mock.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of DataCite.
```
slint: This header is probably from the `inveniosoftware/datacite` module. You can copy one from another Zenodo Python file.

zenodo/modules/records/utils.py
```diff
@@ -27,13 +27,20 @@
...
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
 from invenio_records.api import Record
 from invenio_search import current_search
 from invenio_search.utils import schema_to_index
+from lxml import etree
+from sqlalchemy import or_
 from werkzeug.utils import import_string
 
 from zenodo.modules.openaire import current_openaire
+from zenodo.modules.records.httpretty_mock import httpretty
```
slint: Since this is importing a monkeypatch, just to be safe it would be better to do the import inside the function/task that uses it (ie. inside `xsd41`).

Issue #1444, opened 2018-03-27
Update invenio depdendencies to v1.0.0


slint:
Since most Invenio core packages are now `v1.0.0`, updating one of them on Zenodo requires bumping all other versions to `v1.0.0` as well. Updating packages means fixes + features and maybe behaviour changes on Zenodo. The following packages so far were updated (taken from the [big list of Invenio packages](https://github.com/inveniosoftware/invenio/wiki/Invenio-3.0-Build-Status)):

- `invenio-rest v1.0.0` 
    - CORS installed by default
    - [x] **TODO:** Sentry ID in REST error responses as `error_id` (should be used in Deposit forms and Search results error display :tada:)
- `invenio-records-rest v1.0.1`
    - Datacite v4.1 :tada: :taco:
    - Unicode doesn't change sorting now (closes #1227)
    - removes `invenio-query-parser` dependency (probably safe for us to remove as well)
- `invenio-pidstore v1.0.0`
    - introduces `PIDSTORE_RECID_FIELD`, which doesn't affect us though
- `invenio-search-ui v1.0.1`
    - [x] **TODO:** Fix facet to display selections that don't have any results (addressing inveniosoftware/invenio-search-js#111)
- `invenio-oauthclient v1.0.0`
    - [ ] **TODO:** Test alembic migration for tables to add Timestamp columns
        - ~~:warning: **Important** For zero-downtime, there's a need for an intermediary migration step, which means the existing recipe has to be split into two recipes...~~
    - [x] **TODO:** Check if redirects work as expected (https://github.com/inveniosoftware/invenio-oauthclient/commit/2d6256ed2ac729f6c679265ec0c3053ec7298383). *Note: for some reason the redirect from ORCiD goes to `http` instead of `https`...*
    - Remote access tokens are no longer displayed fully in plaintext
- `invenio-oauth2server v1.0.0`
    - Fixes personal token name's length validation in settings form
- `invenio-oaiserver v1.0.0`
    - [x] **TODO:** OAI-PMH XSL has to be set manually via `OAISERVER_XSL_URL` and be part of Zenodo's static files (https://github.com/inveniosoftware/invenio-oaiserver/commit/112d4248c72535bdf17638e6517d890441ca5197)
- `invenio-pidrelations v1.0.0a4`
    - [x] **TODO:** Compare DB schemas and create alembic stamp for the recipe
- `Jinja v2.10`
[...]

PR #1445, opened 2018-03-28
Invenio + other version bumps

slint:


slint:
List of changes tracked at https://github.com/zenodo/zenodo/issues/1444

PR #1455, opened 2018-04-10
datacite: add export format for v4.1

ChiaraBi:


zenodo/config.py
```diff
@@ -613,13 +613,18 @@ def _(x):
...
     ),
     'dcite3': dict(
         title='DataCite XML',
         serializer='zenodo.modules.records.serializers.datacite_v31',
         order=3,
     ),
+    'dcite4': dict(
+        title='DataCite XML',
+        serializer='zenodo.modules.records.serializers.datacite_v41',
+        order=3,
```
slint: I can't find a Jinja template or some code where this is used... The [box render template](https://github.com/zenodo/zenodo/blob/03dee2c29297da11964bea3fed53d0bb6a1a75dc/zenodo/modules/records/templates/zenodo_records/box/export.html) is explicitly ordering things. 

@lnielsen If there isn't some cryptic place that uses this I would say we remove the `order` key from all of the dicts.

PR #1479, opened 2018-05-24
Fix typo in related identifiers relations

nuest:


slint:
Fixed in https://github.com/zenodo/zenodo/commit/8ea0ad0b8a6db8329137ad5ce69e475ec1a658a1

PR #1516, opened 2018-06-28
stats: add registration for the new queries

ChiaraBi:


zenodo/modules/stats/registrations.py
```diff
@@ -121,3 +122,102 @@ def register_aggregations():
...
+            query_class=ESTermsQuery,
+            query_config=dict(
+                index='stats-file-download',
+                doc_type='file-download-day-aggregation',
+                copy_fields=dict(
+                    conceptrecid='conceptrecid',
+                    conceptdoi='conceptdoi',
+                    communities=lambda d, _: (list(d.communities)
+                                              if d.communities else None),
+                    is_parent=lambda *_: True
```
slint: Hm, we need top implement a `default_filter` on the `*-all-versions` queries where we can specify `is_parent:true`

zenodo/modules/stats/registrations.py
```diff
@@ -148,6 +148,7 @@ def register_queries():
                     recid='recid',
                 ),
                 metric_fields=dict(
+                    count=('sum', 'count'),
```
slint: I removed the default `total` (which was just a sum metric for the `count` field) and made it part of the `metric_fields` parameter. So now we have to basically explicitly add it.

zenodo/modules/stats/registrations.py
```diff
@@ -167,12 +168,13 @@ def register_queries():
                     is_parent='is_parent'
                 ),
                 query_modifiers=[
-                    lambda query, _: query.filter('term', is_parent=True)
+                    lambda query, **_: query.filter('term', is_parent=True)
```
slint: The actual call looked something like:

```python
for modifier in self.query_modifiers:
    self.agg_query = modifier(self.agg_query, **kwargs)
```

so "kwargs" has to be star-expanded

PR #1535, opened 2018-07-11
Tests and fixes for stats

slint:


slint:
We can wait for https://github.com/inveniosoftware/invenio-stats/pull/71 + some more changes before merging

PR #1538, opened 2018-07-16
records: add stats to the record json serializer

### LINKED ISSUE #52: stats: expose metrics via records rest api

ChiaraBi:
* close #52

### DIFF
--- a/zenodo/modules/records/serializers/schemas/json.py
+++ b/zenodo/modules/records/serializers/schemas/json.py
@@ -33,7 +33,7 @@ from marshmallow import Schema, ValidationError, fields, missing, \
 from werkzeug.routing import BuildError
 
 from zenodo.modules.records.utils import is_deposit
-from zenodo.modules.stats.utils import build_stats
+from zenodo.modules.stats.utils import get_record_stats
 
 from . import common
 from ...models import AccessRight, ObjectType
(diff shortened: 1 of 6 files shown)


zenodo/modules/records/serializers/schemas/json.py
```diff
@@ -248,6 +249,16 @@ class RecordSchemaV1(common.CommonRecordSchemaV1):
     revision = fields.Integer(dump_only=True)
     updated = fields.Str(dump_only=True)
 
+    _stats = fields.Method('dump_stats')
```
slint: plain `stats` is fine as well (in the same way we have `files`)

zenodo/modules/stats/utils.py
```diff
@@ -50,3 +51,52 @@ def extract_event_record_metadata(record):
         communities=record.get('communities'),
         owners=record.get('owners'),
     )
+
+
+def build_stats(recid, conceptrecid):
```
slint: Maybe rename this one to `build_record_stats`

zenodo/modules/records/serializers/schemas/json.py
```diff
@@ -248,6 +249,16 @@ class RecordSchemaV1(common.CommonRecordSchemaV1):
...
+
+    def dump_stats(self, obj):
+        """Dump the stats to a dictionary."""
+        if '_stats' in obj:
+            return obj['_stats']
+        else:
+            return build_stats(obj['metadata']['recid'],
```
slint: BTW a nice extra test worth adding is:

1. Mock `build_stats`
2. Perform a Search API request
3. Checking if the mock was called [via `.called`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.called)

PR #1550, opened 2018-07-18
search-ui: change sort option name

ChiaraBi:


zenodo/config.py
```diff
@@ -816,9 +816,9 @@ def _(x):
             default_order='asc',
             order=1,
         ),
-        mostviews=dict(
+        mostviewed=dict(
```
lnielsen: reminder: we will have to update links in FAQ and blog post

PR #1553, opened 2018-07-20
docs: add empty architecure section

ChiaraBi:


docs/architecture/architecture.rst
```diff
@@ -0,0 +1,92 @@
...
+    In applying this license, CERN does not
+    waive the privileges and immunities granted to it by virtue of its status
+    as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+
+=============
+ Architecture
+=============
+
+Statistics
```
slint: Since we're going to add more sections under architecture it would be good to separate them per file and end up with the following directory structure:

```
docs/architecture
├── statistics.rst
├── github.rst
├── communities.rst
├── versioning.rst
└── ...
```

docs/architecture/architecture.rst
```diff
@@ -0,0 +1,92 @@
...
+    * ``/record/<recid>``
+    * ``/record/<recid>/export/<format>``
+
+* ``file-download``: this event is related to the download a file of a record and it's is bounded to
+  the  ``file_downloaded`` signal, which is emitted when one the following endpoints is accessed:
+
+    * ``/record/<recid>/files/<filename>``
+    * ``/api/record/<recid>/files/<filename>``
+
+Every time a user views a record or downloads a file, the corresponding signal is emitted and the corresponding
```
slint: It's worth mentioning that in case an event fails to be emitted (e.g. an exception is raised in an event builder or sending the message to RabbitMQ fails), the request is not affected. A warning is logged though so that we can examine later e.g. in Sentry

docs/architecture/architecture.rst
```diff
@@ -0,0 +1,92 @@
...
+
+* pid_type
+* pid_value
+
+After an event is created, it is sent to a dedicated RabbitMQ queue. Each type of event has its own queue, so
+different types of events can be processed separately.
+The events which are in the queue are then consumed and processed by a Celery task. In this step the
+flags ``is_machine`` and ``is_robot`` are added, the user is anonymized and the double-clicks are deleted. During the data
+anonymization process the user related fields are deleted and replaced by the ``visitor_id`` and the ``unique_session_id``.
+Finally, the processed events are saved in Elasticsearch.
```
slint: Two things worth mentioning here:

- Robot events are discarded for space savings
- In case an event fails to be processed (e.g. malformed IP address), the event is skipped/lost. Same as above, a warning is logged

Issue #1524, opened 2018-07-02
openaire: expose community identifier in DC and DataCite export formats


lnielsen:
The OpenAIRE community (RCD) should be exposed in the two export formats.

```xml
<relatedIdentifier relationType=“isPartOf” relatedIdentifierType=“URL”>http://openaire.eu/communities/fam/</relatedIdentifier>



<dc:relation>http://openaire.eu/communities/fam/</dc:relation>
```

PR #1557, opened 2018-07-24
schemas: change identifier for foreign DOIs

### LINKED ISSUE #1327: openaire: foreign DOIs in export formats
Records which have a foreign DOI (i.e. non-Zenodo DOI) needs to have the export format updated.

The DOI should be moved from <identifier> to relatedIdentifier (type=isIdenticalTo) and use Zenodo URL in identifier.

https://issue.openaire.research-infrastructures.eu/issues/629

slint:
closes #1327 and #1524

### DIFF
--- a/zenodo/modules/openaire/helpers.py
+++ b/zenodo/modules/openaire/helpers.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Zenodo.
-# Copyright (C) 2017 CERN.
+# Copyright (C) 2017-2018 CERN.
 #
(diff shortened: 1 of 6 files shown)


zenodo/modules/openaire/helpers.py
```diff
@@ -122,3 +124,20 @@ def openaire_link(record):
             oaid,
         )
     return None
+
+
+def resolve_openaire_communities(communities):
+    """Resolve a list of Zenodo communities a set of OpenAIRE communities."""
```
ChiaraBi: "Resolve a list of Zenodo communities as a set of OpenAIRE communities."

PR #1559, opened 2018-07-25
Deposit API validation fixes

slint:


tests/unit/conftest.py
```diff
@@ -748,24 +748,38 @@ def funder_record(db):
 
 
 @pytest.fixture
-def grant_record(db, funder_record):
+def grant_records(db, funder_record):
     """Create a grant record."""
```
ChiaraBi: Maybe it's better to change the doc string to "Create grant records"

Issue #414, opened 2015-11-17
stats: send usage events using piwik tracker api


lnielsen:
- OpenAIRE issue: https://issue.openaire.research-infrastructures.eu/issues/1671
- Tracking API: https://developer.piwik.org/api-reference/tracking-api

**Piwik Bulk API Request**

| Parameter | Example | Comment |
| --------- | ------- | ------- |
| `idsite` | "your-site-id" | PiwikID |
| `rec` | 1 | mandatory, doesn't mean anything |
| `url` | `https://zenodo.org/record/5` | URL of the record |
| `download` | `https://zenodo.org/record/5/files/article.pdf` | if it's a `file-download` event, same as `url` |
| `_id` | `bbbbbeef12345678` | `visitor_id` |
| `cid` | `bbbbbeef12345678` | `visitor_id` |
| `cvar` | `{ "1": [ "oaipmhID", "oai:zenodo.org:5" ] }` | `_oai.id` |
| `country` | `fr` | `country.lower()` |
| `cdt` | `2018-01-01 12:00:00` or `1301919102` | `timestamp` |
| `urlref` | google.com | `referer` |
| `action_name` | "A study on astrophysics" | title of the record |

```python
# POST https://analytics.openaire.eu/piwik.php

{
    # List of querystrings with above parameters for each event
    "requests": [
        "?idsite=<your-site-id>&rec=1&url=https://sandbox.zenodo.org/record/5&_id=bbbbbeef12345678&cvar={\"1\":[\"oaipmhID\",\"oai:zenodo.org:5\"]}&...",
        "?idsite=<your-site-id>&rec=1&url=https://sandbox.zenodo.org/record/6&_id=aaaabeef12345678&cvar={\"1\":[\"oaipmhID\",\"oai:zenodo.org:6\"]}&...",
        ...,
    ],
    "token_auth": "<your-api-key>"
}
```

A scheduled task can send events directly queried incrementally (by keeping a "bookmark") from the `events-stats-record-view`/`events-stats-file-download` ES indices. 

**API**

```python
# zenodo.modules.stats.exporters
class PiwikExporter:
    def run(self, start_date, end_date):
        ...

# in the future...
class DataciteExporter:
    def run(self, start_date, end_date):
        ...
```

lnielsen:
Should rely on http://github.com/lnielsen/invenio-stats

PR #1560, opened 2018-07-27
stats: add PiwikExporter

ChiaraBi:
* address #414

### DIFF
--- a/tests/unit/stats/test_stats_exporters.py
+++ b/tests/unit/stats/test_stats_exporters.py
@@ -1,0 +1,131 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Zenodo.
+# Copyright (C) 2018 CERN.
+#
+# Zenodo is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Zenodo is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Zenodo. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Unit tests for statistics exporters."""
+from datetime import datetime, timedelta
+
+import pytest
+from invenio_cache import current_cache
+from mock import mock
+from stats_helpers import create_stats_fixtures
+
+from zenodo.modules.stats.exporters import PiwikExporter, \
+    PiwikExportRequestError
+
+
+class MockResponse:
+    def __init__(self, json_data, status_code):
+        self.json_data = json_data
+        self.status_code = status_code
+        self.ok = status_code == 200
+
+    def json(self):
+        return self.json_data
+
+
+def mocked_requests_success(*args, **kwargs):
+    json = {'status': 'success', 'invalid': 0}
+    return MockResponse(json, 200)
+
+
+def mocked_requests_invalid(*args, **kwargs):
+    json = {'status': 'success', 'invalid': 1}
+    return MockResponse(json, 200)
+
+
+def mocked_requests_fail(*args, **kwargs):
+    return MockResponse({}, 500)
+
+
+@mock.patch('zenodo.modules.stats.exporters.requests.post',
+            side_effect=mocked_requests_success)
+def test_piwik_exporter(app, db, es, locations, event_queues, full_record):
+    records = create_stats_fixtures(
+        metadata=full_record, n_records=1, n_versions=1, n_files=1,
+        event_data={'user_id': '1', 'country': 'CH'},
+        # 4 event timestamps
+        start_date=datetime(2018, 1, 1, 13),
+        end_date=datetime(2018, 1, 1, 15),
+        interval=timedelta(minutes=30),
+        do_process_events=True,
+        do_aggregate_events=False,
+        do_update_record_statistics=False
+    )
+
+    current_cache.delete('piwik_export:bookmark')
+    bookmark = current_cache.get('piwik_export:bookmark')
+    assert bookmark is None
+
+    start_date = datetime(2018, 1, 1, 12)
+    end_date = datetime(2018, 1, 1, 14)
+    PiwikExporter().run(start_date=start_date, end_date=end_date)
+    bookmark = current_cache.get('piwik_export:bookmark')
+    assert bookmark == u'2018-01-01T14:00:00'
+
+    PiwikExporter().run()
+    bookmark = current_cache.get('piwik_export:bookmark')
+    assert bookmark == u'2018-01-01T14:30:00'
+
+
+@mock.patch('zenodo.modules.stats.exporters.requests.post',
+            side_effect=mocked_requests_invalid)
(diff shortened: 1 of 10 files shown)


zenodo/modules/stats/exporters.py
```diff
@@ -0,0 +1,135 @@
...
+            'range', timestamp=time_range
+        ).sort(
+            {'timestamp': {'order': 'asc'}}
+        ).params(preserve_order=True).scan()
+
+        url = current_app.config['ZENODO_STATS_PIWIK_EXPORTER'].get('url', None)
+        token_auth = current_app.config['ZENODO_STATS_PIWIK_EXPORTER'] \
+            .get('token_auth', None)
+
+        chunk_size = 85
```
slint: Maybe add a comment here explaining why it's 85 (i.e. `[max piwik payload size = 64k] / [max querystring size = 750]`. Even better you can move it in the `ZENODO_STATS_PIWIK_EXPORTER` config (including the comment), so that we can change it in case we see any troubles.

zenodo/modules/stats/exporters.py
```diff
@@ -0,0 +1,135 @@
...
+            .get('id_site', None)
+        url = 'zenodo.org/record/{}'.format(event.recid)
+        visitor_id = event.visitor_id[0:16]
+        oai = record.get('_oai').get('id')
+        cvar = "{\"1\":[\"oaipmhID\",\"" + oai + "\"]}"
+        action_name = record.get('title')[:150]  # max 150 characters
+
+        if 'file_key' in event:
+            url = '{}/files/{}'.format(url, event.file_key)
+            query_string = '?idsite={}&rec=1&url={}&_id={}&cid={}&cvar={}&' \
```
slint: To keep it safe in terms of URL-encoding, you can use [urllib.parse.urlencode](https://docs.python.org/3/library/urllib.request.html#urllib-examples) and also simplify the file-specific logic by using a dictionary with the values. Unfortunately, since we're targeting both Python 2 + 3 you have to use `from six.moves.urllib.parse import urlencode` :sob:

zenodo/modules/stats/config.py
```diff
@@ -0,0 +1,32 @@
...
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Configuration for Zenodo stats."""
+
+ZENODO_STATS_PIWIK_EXPORTER = {
+    'id_site': 17,
```
slint: I guess it's better to set this one to `1`, which is the test website

zenodo/modules/stats/exporters.py
```diff
@@ -0,0 +1,135 @@
...
+                    current_app.logger.warning(msg, extra=extra)
+                elif update_bookmark is True:
+                    current_cache.set('piwik_export:bookmark',
+                                      event_chunk[len(event_chunk)-1].timestamp,
+                                      timeout=-1)
+            else:
+                msg = 'Invalid events in Piwik export.'
+                extra = {
+                    'begin_event': event_chunk[0].timestamp,
+                    'end_event': event_chunk[len(event_chunk) - 1].timestamp
```
slint: `event_chunk[-1] == event_chunk[len(event_chunk) - 1]`

zenodo/modules/stats/exporters.py
```diff
@@ -0,0 +1,135 @@
...
+
+            res = requests.post(url, json=payload)
+
+            # Failure: not 200 or not "success"
+            content = json.loads(res.content)
+            if res.status_code == 200 and content.get('status') == 'success':
+                if content.get('invalid') != 0:
+                    msg = 'Invalid events in Piwik export.'
+                    extra = {
+                        'begin_event': event_chunk[0].timestamp,
```
slint: `begin_event_timestamp` is a bit more descriptive

zenodo/modules/stats/exporters.py
```diff
@@ -0,0 +1,135 @@
...
+                                      event_chunk[len(event_chunk)-1].timestamp,
+                                      timeout=-1)
+            else:
+                msg = 'Invalid events in Piwik export.'
+                extra = {
+                    'begin_event': event_chunk[0].timestamp,
+                    'end_event': event_chunk[len(event_chunk) - 1].timestamp
+                }
+                current_app.logger.warning(msg, extra=extra)
+                raise Exception()
```
slint: Since an exception is raised anyways, you can create a custom exception (like e.g. [MissingCommunityError](https://github.com/zenodo/zenodo/blob/9f0c975184f5f59e2aaf503134ceec8873b29a3e/zenodo/modules/deposit/errors.py#L64:L72)) and include the `msg` and `extra` dictionary

zenodo/modules/stats/exporters.py
```diff
@@ -0,0 +1,136 @@
...
+
+            payload = {
+                'requests': query_strings,
+                'token_auth': token_auth
+            }
+
+            res = requests.post(url, json=payload)
+
+            # Failure: not 200 or not "success"
+            content = json.loads(res.content) if res.ok else None
```
slint: `json.loads(res.content) -> res.json()`

tests/unit/stats/test_stats_exporters.py
```diff
@@ -0,0 +1,144 @@
...
+    def json(self):
+        return self.json_data
+
+
+def mocked_requests_success(*args, **kwargs):
+    url = current_app.config['ZENODO_STATS_PIWIK_EXPORTER'].get('url', None)
+    json = {'status': 'success', 'invalid': 0}
+
+    if args[0] == url:
+        return MockResponse(json, 200)
```
slint: I don't think you need to check for the URL, since all of the calls inside the test will be towards the same one always.

tests/unit/stats/test_stats_exporters.py
```diff
@@ -0,0 +1,144 @@
...
+def test_piwik_exporter_invalid_request(app, db, es, locations, event_queues,
+                                        full_record):
+    records = create_stats_fixtures(
+        metadata=full_record, n_records=1, n_versions=1, n_files=1,
+        event_data={'user_id': '1', 'country': 'CH'},
+        # 4 event timestamps
+        start_date=datetime(2018, 1, 1, 13),
+        end_date=datetime(2018, 1, 1, 15),
+        interval=timedelta(minutes=30),
+        do_process_events=True)
```
slint: You can skip the rest of the fixture creation steps by passing `do_aggregate_events=False` and `do_update_record_statistics=False`, to save on some test duration

zenodo/modules/stats/exporters.py
```diff
@@ -103,7 +103,7 @@ def run(self, start_date=None, end_date=None, update_bookmark=True):
                     'begin_event_timestamp': event_chunk[0].timestamp,
                     'end_event_timestamp': event_chunk[-1].timestamp,
                 }
-                raise PiwikExportRequestError(msg, export_info=info)
+                raise PiwikExportRequestError(info, msg)
```
slint: It's better to be explicit and pass the info as a keyword parameter (i.e. `export_info=info`), since in the constructor of the `PiwikExportRequestError` it's "consumed" as such.

zenodo/modules/stats/exporters.py
```diff
@@ -0,0 +1,136 @@
...
+                }
+                raise PiwikExportRequestError(msg, export_info=info)
+
+    def _build_query_string(self, event):
+        id_site = current_app.config['ZENODO_STATS_PIWIK_EXPORTER']\
+            .get('id_site', None)
+        url = 'zenodo.org/record/{}'.format(event.recid)
+        visitor_id = event.visitor_id[0:16]
+        _, record = fetch_record(event.recid)
+        oai = record.get('_oai').get('id')
```
slint: record.get('_oai', {}).get('id')

Issue #738, opened 2016-09-05
openaire: update RCD communities


lnielsen:
Phase 1:
- Create community for openaire EGI community
- Export community identifier in DC\

Phase 2:
- Reserve namespace for openaire communities
- Export community creation via rest api

krzysztof:
@lnielsen Requires better explanation

dnet-team:
Was this for us?

Lars, let us know if you need any clarification. 

Paolo

----
Paolo Manghi
Istituto di Scienza e Tecnologie dell'Informazione "A. Faedo" - CNR
Tel: +39 050 315-2038 - Fax: +39 050 315-3464
Web: http://www.isti.cnr.it/People/P.Manghi
InfraScience Research Group: http://nemis.isti.cnr.it/groups/infrascience

lnielsen:
@dnet-team No it's for me :-)

alessiabardi:
Hi Lars,

please find attached the list of openaire-connect communities.
The file is a csv with three columns:
1. communityid: string that identifies a community in OpenAIRE. This is the value to use to link a product to an OpenAIRE community.
2. communityname: the name of the community (for display purposes)
3. zenodo communities: a ';' separated list of zenodo communities that can be mapped into the OpenAIRE community. The mapping is of course preliminary (I made a very basic search among the Zenodo community). If you know about other communities that should be mapped, feel free to enrich the csv and let me know.

If you think it is worth it, feel free to add the csv to github. 

[openaire-zenodo_communities_2017-10-24.csv.txt](https://github.com/zenodo/zenodo/files/1411332/openaire-zenodo_communities_2017-10-24.csv.txt)

paolomanghi:
We should add a prefix to the communities above, of the likes:

OpenAIRE-Connect - <community name>

Any other suggestion is welcome.

krzysztof:
Mapping translation from the CSV file to Google docs: https://docs.google.com/spreadsheets/d/1L3ZK8hvsgjHF-q50fog3fyZBY7FhF7NzzE83BTDvZJ8/edit#gid=0

mirima2:
Hi all,

To the csv file should be added one entry related to the new communit Instruct-ERIC:

OpenAIRE_ID: eric; 
OpenAIRE_name: Instruct-ERIC; 
Zenodo_ID: Instruct, West-Life Virtual Research Environment for Structural Biology

alessiabardi:
Hi @krzysztof ,

[...]

PR #1566, opened 2018-08-03
utils: cli command to update communities mapping

ChiaraBi:
* addresses: #738

### DIFF
--- a/zenodo/modules/utils/cli.py
+++ b/zenodo/modules/utils/cli.py
@@ -26,6 +26,7 @@
 
 from __future__ import absolute_import, print_function
 
+import json
 import os
 from io import SEEK_END, SEEK_SET
 
@@ -42,6 +43,7 @@ from zenodo.modules.deposit.tasks import datacite_register
 from zenodo.modules.records.resolvers import record_resolver
 
 from .grants import OpenAIREGrantsDump
+from .openaire import OpenAIRECommunitiesMappingUpdater
 from .tasks import has_corrupted_files_meta, repair_record_metadata, \
     sync_record_oai, update_oaisets_cache, update_search_pattern_sets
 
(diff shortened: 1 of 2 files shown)


zenodo/modules/utils/communities.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-
```
slint: Since this is more OpenAIRE-specific, `openaire.py` would be a better name for the file

zenodo/modules/utils/communities.py
```diff
@@ -0,0 +1,78 @@
...
+            for zenodo_community in openaire_community.get('communities'):
+                # search for the corresponding community in Zenodo
+                query = db.session.query(Community).filter(
+                    Community.title.like(zenodo_community + '%')
+                )
+                if query.count() == 1:
+                    comm = query.one()
+                    zenodo_community_ids.append(comm.id)
+                else:
+                    print('Community not found:\n' + zenodo_community)
```
slint: Instead of printing here, you can return as the result of the method two items:

```python
...
def update_communities_mapping(self)
    mapping = ...
    unresolved_communities = []
    # ...update the "mapping" and keep track of any unresolved communities...
    # if error:
    #     unresolved_communities.append(zenodo_community)
    return mapping, unresolved_communities
```

And then you can print the unresolved communities using `click.echo` in `cli.py`

zenodo/modules/utils/communities.py
```diff
@@ -0,0 +1,78 @@
...
+        """Update communities mapping."""
+        mapping = current_app.config['ZENODO_OPENAIRE_COMMUNITIES']
+        for openaire_community in self._communities_iterator:
+            zenodo_community_ids = []
+            for zenodo_community in openaire_community.get('communities'):
+                # search for the corresponding community in Zenodo
+                query = db.session.query(Community).filter(
+                    Community.title.like(zenodo_community + '%')
+                )
+                if query.count() == 1:
```
slint: You can also do `comm = query.one_or_none()`

PR #1621, opened 2018-10-12
Escape additional special characters in BibTex serializer

Xarthisius:
It's #997 again. Underscore is a fairly common char in python packages (see e.g. https://zenodo.org/record/1458961#.W7_VIXFKhTY). This PR adds additional characters that should be escaped when serializing to BibTex. Please note that set is still incomplete. Specifically we're missing: `~`, `^`, `\`, which should be mapped to `\textasciitilde{}`, `\textasciicircum{}`, `\textbackslash{}` respectively and possibly others.


slint:
Closed in favor of #1922 (sorry @Xarthisius, just saw that a relevant PR already existed)

PR #1625, opened 2018-10-19
stats: aggregations fault tolerance increase

slint:
* Uses a custom Elastricsearch client with increased request timeout
  when processing aggregations.
* Decreases days batch size for aggregation queries (creating buckets
  for 1 day vs. 7 days).

### DIFF
--- a/tests/unit/stats/test_stats_aggs.py
+++ b/tests/unit/stats/test_stats_aggs.py
@@ -90,8 +90,12 @@ def test_large_stats(app, db, es, locations, event_queues, minimal_record):
 
     # Aggregations indices
     # (4 versions + 1 concept) * 3 records -> 15 documents + 2 bookmarks
-    assert search.index('stats-file-download').count() == 933  # 2bm + 30d
-    assert search.index('stats-record-view').count() == 933  # 2bm + 30d
+    q = search.index('stats-file-download')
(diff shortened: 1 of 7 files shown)


tests/unit/stats/test_stats_aggs.py
```diff
@@ -90,8 +90,12 @@ def test_large_stats(app, db, es, locations, event_queues, minimal_record):
 
     # Aggregations indices
     # (4 versions + 1 concept) * 3 records -> 15 documents + 2 bookmarks
-    assert search.index('stats-file-download').count() == 933  # 2bm + 30d
-    assert search.index('stats-record-view').count() == 933  # 2bm + 30d
+    q = search.index('stats-file-download')
+    q = q.doc_type('file-download-day-aggregation')
+    assert q.count() == 915  # 61 days * 15 records
```
slint: The reason the total number of documents changed, was because there are more bookmarks created (since the batches are more often, i.e. everyday). To avoid a regression of this failure, the document type is also included in the query explicitly now.

slint: The reason the total number of documents changed, was because there are more bookmarks created (since the batches are more often, i.e. everyday). To avoid a regression of this failure, the document type is also included in the query explicitly now.

PR #1627, opened 2018-10-22
stats: don't export events if bookmark is None

ChiaraBi:


zenodo/modules/stats/exporters.py
```diff
@@ -47,6 +47,10 @@ def run(self, start_date=None, end_date=None, update_bookmark=True):
         """Run export job."""
         if start_date is None:
             bookmark = current_cache.get('piwik_export:bookmark')
+            if bookmark is None:
+                msg = 'Bookmark not found.'
```
slint:
```suggestion
                msg = 'Bookmark not found, and no start date specified.'
```

PR #1632, opened 2018-10-30
Add reviews/isReviewedBy to deposit form and render

stain:
example of a review with DOI from PeerJ: 
https://doi.org/10.7287/peerj.5551v0.2/reviews/1

Zenodo should also be able to support linking to peer review. 

is-reviewed-by

Adding this will allow us to update a record to link to DOI and HTTPs of what has reviewed say a preprint; 
alternatively to deposit a peer review report and indicate what paper/preprint it is reviewing.

Proper Open Science! :)

See DataCite relations:
https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/data/include/v4.1/datacite-relationType-v4.1.xsd#L34

(perhaps others of these should also be considered)

I notice this is already allowed in the schemas:

```
./zenodo/modules/records/jsonschemas/records/base-v1.0.0.json:        "isReviewedBy",
./zenodo/modules/records/jsonschemas/records/record-v1.0.0.json:        "isReviewedBy",
./zenodo/modules/records/jsonschemas/records/record-v1.0.0.json:              "isReviewedBy",
./zenodo/modules/deposit/jsonschemas/deposits/records/record-v1.0.0.json:        "isReviewedBy",
./zenodo/modules/deposit/jsonschemas/deposits/records/record-v1.0.0.json:              "isReviewedBy",
./zenodo/modules/deposit/jsonschemas/deposits/records/legacyrecord.json:              "isReviewedBy",
```

WIP - how do I verify the relation makes it way into DataCite registration?
Should look similar to (is-review-of) in
http://api.crossref.org/v1/works/10.7287/peerj.5551v0.2/reviews/1


slint:
Clsoing, since these relation types (and others) have been added in #1877

PR #1641, opened 2018-11-16
exporter: add errors and skip deleted records

ChiaraBi:


zenodo/modules/exporter/errors.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class FailedExportJobError(Exception):
+    """Error for failed export job."""
+
+    def __init__(self, record_ids=None):
+        """Initialize the error with the list of not serialized records."""
+        msg = "Serialization failed for the following records: "
+        for r in record_ids:
+            msg = msg + r + ", "
```
slint: You can also do:

```python
msg = "Serialization failed for the following records: {}".format(', '.join(record_ids))
```

PR #1654, opened 2018-12-04
theme: fix citation filters

ChiaraBi:


zenodo/modules/theme/static/js/zenodo/filters/citations.js
```diff
@@ -73,13 +73,13 @@ define([], function() {
...
     return function(relationship) {
       var doi = "";
-
-      relationship.metadata.Source.Identifier.forEach( function(identifier) {
-        if (identifier.ID && identifier.IDScheme == "doi") {
-            doi = identifier.ID;
-        }
-      })
-
+      if(relationship.metadata.Source.Identifier) {
```
slint: Are there any responses without a `.Identifier` field? (if possible PM me an example)

PR #1677, opened 2019-01-15
Facet

ekpustulka:
We have 
1. added a year facet for publication year filtering
screenshot 2019-01-15 at 16 08 00

2. adjusted journal sort to take journal title instead of year (as year is currently not filled)

3. added the display of journal name, volume, issue, page no in search result
screenshot 2019-01-16 at 14 14 01

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,4 @@ docs/_build/
 target/
 
 test.db
+.idea/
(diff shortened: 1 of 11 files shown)


slint:
Included in https://github.com/zenodo/zenodo/pull/2040. The range filter will be reconsidered and added in future versions

PR #1681, opened 2019-01-17
deposit: fix backend accepting formdata with no 'doi' field

Glignos:


zenodo/modules/records/serializers/schemas/common.py
```diff
@@ -267,7 +267,7 @@ class SubjectSchemaV1(IdentifierSchemaV1):
 class CommonMetadataSchemaV1(Schema, StrictKeysMixin, RefResolverMixin):
     """Common metadata schema."""
 
-    doi = DOIField()
+    doi = DOIField(missing= '')
```
slint:
```suggestion
    doi = DOIField(missing='')
```

tests/unit/deposit/test_api_simpleflow.py
```diff
@@ -146,7 +146,7 @@ def test_simple_rest_flow(mocker, api, api_client, db, es,
     # Not allowed to edit drafts
     response = client.put(
         links['self'], data=json.dumps(test_data), headers=auth_headers)
-    assert response.status_code == 403
+    assert response.status_code == 400
```
slint: I think we should keep this test the way it was (i.e. returning a 403 because of the deposit status) by adding the `doi` field in the `test_data` (it should be there anyway)

erinb90:
Hello, could you please make it clear in the [API documentation](https://developers.zenodo.org/#update) that doi is now a required field when updating a deposition? I was struggling for a while wondering why I kept getting a 400 error until I found this PR. Thanks! :smile:

PR #1682, opened 2019-01-17
records-ui: fix citation filter labels

### LINKED ISSUE #1675: records-ui: proper "for" attributes on citation filters labels
Citation filter labels should have the correct [`for` html attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#Attributes), so that when also clicking on the label (i.e. the text part) of a filter its checkbox should be toggled. ([example record](https://zenodo.org/record/11813))

image

Glignos:
* Closes #1675

### DIFF
--- a/zenodo/modules/theme/static/scss/record.scss
+++ b/zenodo/modules/theme/static/scss/record.scss
@@ -187,9 +187,15 @@
         text-decoration: none;
    }
 
-    .type-filter {
+    .filter {
         margin-right: 5px;
         padding-bottom: -5px;
+        label {
+          font-weight: 300;
+        }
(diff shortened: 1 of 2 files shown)


zenodo/modules/theme/static/scss/record.scss
```diff
@@ -179,9 +179,15 @@
...
 
-    .type-filter {
+    .filter {
         margin-right: 5px;
         padding-bottom: -5px;
+        label {
+          font-weight: 300;
+        }
+        input[type=checkbox][disabled] + label{
+          cursor:not-allowed;
```
slint:
```suggestion
          cursor: not-allowed;
```

PR #1685, opened 2019-01-28
record-ui: fix bibtex long value formatting

### LINKED ISSUE #1598: bibtex: title shown twice
For some unknown reason, the title in this BibTex export is displayed twice: https://zenodo.org/record/1409831/export/hx

Probable cause is some issue with the line wrapping logic...

(Ticket ID: 239315)

Glignos:
* Fix the case of values with trailing whitespace causing
  repetition of the value. (closes #1598)

### DIFF
--- a/zenodo/modules/records/serializers/bibtex.py
+++ b/zenodo/modules/records/serializers/bibtex.py
@@ -301,6 +301,8 @@ class Bibtex(object):
 
     def _format_output_row(self, field, value):
         out = ""
+        if isinstance(value, six.string_types):
+            value = value.strip()
         if field == "author":
             if len(value) == 1:
                 out += u"  {0:<12} = {{{1}}},\n".format(field, value[0])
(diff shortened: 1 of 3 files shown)


zenodo/modules/records/serializers/bibtex.py
```diff
@@ -301,6 +301,7 @@ def _fetch_fields(self, req_fields, opt_fields=None, ign_fields=None):
 
     def _format_output_row(self, field, value):
         out = ""
+        field = field.rstrip()
```
slint: Why not plain `.strip()`?

Glignos: The goal for this is to eliminate trailing whitespace. This should be enough to fix our error case.
Plain stripping means we are reformatting the screen and creates a complexity since you shouldn't stip all spaces, unless I'm missing something

PR #1697, opened 2019-02-06
iiif thumbnail precaching

Glignos:


zenodo/modules/thumbnails/cache.py
```diff
@@ -0,0 +1,52 @@
...
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Implements a Redis cache."""
+
+from __future__ import absolute_import
+
+from flask_iiif.cache.redis import ImageRedisCache as ImageCache
+
+
+class ImageRedisCache(ImageCache):
```
slint: Since this is just a specialized Redis cache, I would keep the import name intact (i.e. `from flask_iiif.cache.redis import ImageRedisCache` and change this to:

```suggestion
class FilteredImageRedisCache(ImageRedisCache):
```

zenodo/modules/thumbnails/tasks.py
```diff
@@ -0,0 +1,74 @@
...
+            "lt":  str(current_preprocessing_time)
+        }}}
+    ).filter({
+        "term": {
+            "communities": community
+        }}
+    ).source(include=['_files']))
+    records_files = q.scan()
+    for record_files in records_files:
+        for object_file in record_files.to_dict()['_files']:
```
slint:
```suggestion
        for object_file in record_files.to_dict().get('_files', []):
```

zenodo/modules/thumbnails/tasks.py
```diff
@@ -0,0 +1,74 @@
...
+            if(object_file['type'] not in ['jpg', 'png', 'tif', 'tiff']):
```
slint:
```suggestion
            if(object_file.get('type') not in ['jpg', 'png', 'tif', 'tiff']):
```

zenodo/modules/thumbnails/tasks.py
```diff
@@ -0,0 +1,74 @@
...
+                continue
+            size = '250,'  # flask_iiif doesn't support ! at the moment
```
slint: I think we don't need `!` anyways, the `w,` is fine

tests/unit/conftest.py
```diff
@@ -1130,3 +1131,13 @@ def g_tester_id(app, db):
     )
     db.session.commit()
     return tester.id
+
+
+@pytest.fixture
+def iiif_cache():
```
slint: I think it's better to pass the `app` fixture
```suggestion
def iiif_cache(app):
```

And then:

```python
cache = app.extensions['iiif'].cache
```

so that you use the application config.

zenodo/modules/thumbnails/cache.py
```diff
@@ -0,0 +1,52 @@
...
+    def set(self, key, value, timeout=None):
+        """Cache the object.
+
+        :param key: the object's key
+        :param value: the stored object
+        :type value: `BytesIO` object
+        :param timeout: the cache timeout in seconds
+        """
+        identifier, _, size, _, _ = key.split('/')
+        if size == '250,':
```
slint: I think we should be checking for very specific parameters (i.e. `.../{full}/{250,}/{0}/{default}.{png}`).

E.g. this one would match even rotated images and cache them

tests/unit/conftest.py
```diff
@@ -1130,3 +1131,13 @@ def g_tester_id(app, db):
...
+    """Fixture for iiif chache."""
+    cache = ImageRedisCache()
+    yield cache
+    iiif_keys = [k for k in cache.cache._client.keys() if k.find("iiif:") != -1]
```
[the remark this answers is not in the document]
ChiaraBi: This is failing in python3.5 :( 
`TypeError: startswith first arg must be bytes or a tuple of bytes, not str`

ChiaraBi:
Tests are failing because [here](https://github.com/inveniosoftware/flask-iiif/blob/master/flask_iiif/restful.py#L64) the key gets encoded. We should make a PR to `flask-iiif` to solve this problem.

PR #1701, opened 2019-02-07
deposit-ui: add part value in publication type

Glignos:


zenodo/modules/records/data/objecttypes.json
```diff
@@ -479,6 +480,26 @@
...
+  "$schema": "https://zenodo.org/schemas/records/objecttype-v1.0.0.json",
+  "internal_id": "publication-part",
+  "id": "https://zenodo.org/objecttypes/publication/part#",
+  "title": {
+    "en": "Part"
+  },
+  "title_plural": {
+    "en": "Parts"
+  },
+  "schema.org": "https://schema.org/ImageObject",
```
lnielsen: ``schema.org``, ``datacite``, ``eurepo``, ``openaire`` and ``csl`` attributes have to be updated as well.

zenodo/modules/deposit/static/json/zenodo_deposit/deposit_form.json
```diff
@@ -135,6 +135,10 @@
           {
             "value": "other",
             "name": "Other"
+          },
+          {
+            "value": "part",
```
lnielsen: Note, we're still pending a final decision on the name. Also, we have to ensure that treatments are not sent via direct indexing to OpenAIRE.

slint:
@lnielsen Have we decided on the naming for these types? If not, we should discuss IRL

slint:
After discussion with @tcatapano, the existing `section` publication sub-type should be sufficient. Renaming the display value from `Book section` to just `Section` would help though (CC @lnielsen)

slint:
Closed in favor of the new "Taxonomic treatment" subtype

PR #1702, opened 2019-02-09
Initial switch to local share buttons

nemobis:
* Avoid third-party resources, works with tracker blockers.
* Avoid leaking PII to AddThis, not mentioned in the privacy policy.


nemobis:
The flask-share plugin could be improved to add support for custom URLs and citeulike, but I wonder how much use these see anyway. I'd think that respecting privacy laws takes priority anyway.

nemobis:
Cf. https://www.codyh.com/writing/tracking.html for some context, https://edri.org/edrigramnumber9-16german-dpa-like-facebook/ for some example.

nemobis:
After the [removal of Altmetric badges](https://blog.zenodo.org/2020/07/09/2020-07-09-altmetric-badges/), I think AddThis is the only third-party resource loaded on records which sends personal identifying information overseas.

Glignos:
Thank you for taking the time to make this contribution. After reviewing the solution we decided that it doesn't quite fit our current needs since only a subset of the previously available sharing options is present. So instead we decided to remove the sharing functionality altogether until we figure out how we would like to proceed on the matter.

nemobis:
Glignos, 07/09/20 11:07:
> So instead we decided to remove the sharing functionality altogether until we figure out how we would like to proceed on the matter.

Ok. Let me know which sharing targets are most interesting for Zenodo; I
might be able to send a patch upstream.

nemobis:
> So instead we decided to remove the sharing functionality altogether until we figure out how we would like to proceed on the matter.

When is that going to happen? I see AddThis is still loaded on Zenodo. Instead of waiting for an ideal solution while the users' private data is sent every day to companies which shouldn't have it, I recommend adopting a gradual solution like my patch which preserves 95 % of the value while reducing the cost by several orders of magnitude.

nemobis:
I'm glad to see that with the switch to the new version Zenodo has finally removed AddThis trackers!

PR #1716, opened 2019-03-01
docker: fix docker installation

frankois:
* update the following configs :
    - docker
    - nginx
    - uwsgi

* integrate invenio-files-rest 1.0.0a23.post1


docker-compose.full.yml
```diff
@@ -64,6 +62,12 @@ services:
       service: flower
     links:
       - mq
+  kibana:
+    extends:
+      file: docker-services.yml
+      service: kibana
+    links:
+      - es
```
Glignos: Accidental transfer?

docker-services.yml
```diff
@@ -6,6 +6,8 @@ services:
     environment:
       - "INVENIO_ACCOUNTS_SESSION_REDIS_URL=redis://cache:6379/1"
       - "INVENIO_BROKER_URL=amqp://guest:guest@mq:5672//"
+      - "INVENIO_CELERY_BROKER_URL=amqp://guest:guest@mq:5672//"
+      - "INVENIO_QUEUES_BROKER_URL=amqp://guest:guest@mq:5672//"
```
Glignos: This is not needed since we are doing this https://github.com/zenodo/zenodo/blob/de48b2b81ecaad53481afe30fbbf8fd6b3493974/zenodo/config.py#L1298

docker-compose.yml
```diff
@@ -67,9 +17,10 @@ services:
     extends:
       file: docker-services.yml
       service: es
-  kibana:
+  # Monitoring
+  flower:
```
Glignos: Do we need flower in a simple docker compose?

docker/uwsgi/uwsgi.ini
```diff
@@ -1,7 +1,31 @@
...
+ignore-write-errors = true
+disable-write-exception = true
+
 processes = 2
 threads = 2
+thunder-lock = true
+
+module = zenodo.wsgi
+callable = application
+virtualenv = /usr/local/
```
Glignos: Newline missing

docker/nginx/nginx.conf
```diff
@@ -1,68 +1,51 @@
...
+    types_hash_max_size 2048;
 
-    keepalive_timeout  65;
+    include             /etc/nginx/mime.types;
+    default_type        application/octet-stream;
 
     gzip on;
     gzip_disable "msie6";
     gzip_http_version 1.1;
-    gzip_comp_level 5; # or anything between 4-6
```
Glignos: Is this comment misleading?

docker/nginx/conf.d/default.conf
```diff
@@ -27,35 +29,38 @@ server {
...
+    uwsgi_param X-Request-ID $request_id;
+    client_max_body_size 1024G;
   }
   location /static {
     alias /usr/local/var/instance/static;
+    add_header Cache-Control "max-age=0, must-revalidate";
     autoindex off;
   }
-}
+}
```
Glignos: Newline missing

requirements.txt
```diff
@@ -117,7 +117,7 @@ invenio-config==1.0.0
 invenio-csl-rest==1.0.0a1
 invenio-db==1.0.0
 invenio-deposit==1.0.0a9
-invenio-files-rest==1.0.0a23
+invenio-files-rest==1.0.0a23.post1
```
Glignos: Requirements.txt should be kept on par with setup.py, so implement the change there as well :)

docker-compose.yml
slint: It's sometimes useful to be able to see what's running (for long-running tasks, or failed ones)

docker-compose.full.yml
frankois: No, this was moved out of the 'base services' section into the 'monitoring' section to be more consistent

docker/nginx/nginx.conf
frankois: No indeed removing it was not intended, I just added it back

PR #1720, opened 2019-03-07
records-ui: remove citation box if no doi field

### LINKED ISSUE #1694: records-ui: Do not display citation box with display without DOI
Example : /record/7032

Glignos:
* Closes #1694


zenodo/modules/records/templates/zenodo_records/box/citations.html
```diff
@@ -22,6 +22,7 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 -#}
 
+{% if record.doi %}
 {% set show_betafeature = 'citations' in session.get('featureFlags', []) or request.args.get('feature') == 'citations' %}
 {% set show_citations = current_user.has_role('admin') or config['ZENODO_RECORDS_UI_CITATIONS_ENABLE'] %}
```
slint: To avoid the extra level of nesting, maybe include the check in this `show_citations` variable

Glignos: One of them can still be true though. 
Since the case is that we don't want in any case to load it in the template, I transferred it into the final condition.

PR #1721, opened 2019-03-07
deposit-ui: fix redirect link generated

### LINKED ISSUE #1622: deposit: error while trying to create a new version
In order to reproduce the problem, follow the steps below:
1. Go to a record page;
2. Click on the "Edit" button;
3. Click on "New version" in the deposit form;
4. Click on "Create new version".

Glignos:
* Fixes redirect link for ui  provided by the api
  when creating a new version (closes #1622).


Glignos:
This is only fixing the problem partially, since we also need to tackle the issue of both actions running simultaneously and causing an elastic search version conflict.

PR #1730, opened 2019-03-18
upgrade: bump packages to the latest version

frankois:
* run isort -y
* update manifest


setup.py
```diff
@@ -146,10 +145,10 @@
     'invenio-records-ui>=1.0.0',
     'invenio-records>=1.0.0',
```
slint: Minimum versions of packages should also be bumped (if they were bumped in `requirements.txt`)

PR #1747, opened 2019-04-04
deposit: support temporal metadata in REST APIs

### LINKED ISSUE #1741: deposit: temporal metadata REST API
**Data model**

```python
{
   "dates": [
      {"start": "2019-03-31", "end": "2019-04-01", "type": "collected", "description": "A freeform textual description"},
   ],
}
```

**Deposit API**

- Granularity should be to the level of date: `YYYY-MM-DD`
- `type`: Controlled vocabulary from DataCite (to check if all or only some are made available)
- Validation: `start` should be before `end`, it should be possible to provide only `start` or only `end`, there should be at least one of them
- For single/exact values should be implemented by setting `start` and `end` to the same value

### LINKED ISSUE #1742: search: temporal metadata
- ES mappings: ensure proper searching can be done
  - Proper date granularity type
  - `object` type should be used. We might later migrate to `nested`

### LINKED ISSUE #1743: serializers: temporal metadata
**Serialization: DataCite XML**

```xml
<dates>
<date dateType="collected" dateInformation="....">2017-09-13/2017-10-13</date>
</dates>
```

**Searlization**

- [x] JSON
- [x] DataCite
- [x] Dublin Core
- [x] JSON-LD
- [ ] ~~MARC21~~ https://github.com/zenodo/zenodo/issues/1760
- ... ?

ChiaraBi:
* support temporal metadata in REST APIs (close #1741)
* serializers for temporal metadata (close #1743)
* support temporal search queries (close #1742)

### DIFF
--- a/zenodo/modules/records/serializers/schemas/common.py
+++ b/zenodo/modules/records/serializers/schemas/common.py
@@ -264,6 +264,15 @@ class SubjectSchemaV1(IdentifierSchemaV1):
     term = SanitizedUnicode()
 
 
+class DateSchemaV1(Schema):
+    """Schema for date intervals."""
+
+    start = DateString()
+    end = DateString()
+    type = fields.Str(required=True)
+    description = fields.Str()
+
+
 class CommonMetadataSchemaV1(Schema, StrictKeysMixin, RefResolverMixin):
     """Common metadata schema."""
 
@@ -272,6 +281,8 @@ class CommonMetadataSchemaV1(Schema, StrictKeysMixin, RefResolverMixin):
     title = SanitizedUnicode(required=True, validate=validate.Length(min=3))
     creators = fields.Nested(
         PersonSchemaV1, many=True, validate=validate.Length(min=1))
+    dates = fields.List(
+        fields.Nested(DateSchemaV1), validate=validate.Length(min=1))
     description = SanitizedHTML(
         required=True, validate=validate.Length(min=3))
     keywords = fields.List(SanitizedUnicode())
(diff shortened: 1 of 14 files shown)


zenodo/modules/records/serializers/schemas/common.py
```diff
@@ -305,6 +315,26 @@ def validate_language(self, value):
                 field_name=['language']
             )
 
+    @validates('dates')
+    def validate_dates(self, value):
+        """Validate that start date is before the corresponding end date."""
+        for interval in value:
+            start = arrow.get(interval.get('start')).date() \
```
slint: Arrow is a bit "liberal" with parsing stuff, so it would be better to be more explicit/strict about the calls:

```python
import arrow

arrow.get(None)
# <Arrow [2019-04-04T13:52:47.838153+00:00]>
arrow.get('2019')  # will do weird unix timestamp parsing...
# <Arrow [1970-01-01T00:33:39+00:00]>
arrow.get('2019-03')  # not necessarily bad...
# <Arrow [2019-03-01T00:00:00+00:00]>

# Strict format:
arrow.get(None, 'YYYY-MM-DD')
# TypeError: Can't parse two arguments of types '<type 'NoneType'>', '<type 'str'>' 
arrow.get('2019', 'YYYY-MM-DD')
# ParserError: Failed to match '(?P<YYYY>\d{4})\-(?P<MM>\d{2})\-(?P<DD>\d{2})' when parsing '2019'
arrow.get('2019-03', 'YYYY-MM-DD')  # we lost this...
# ParserError: Failed to match '(?P<YYYY>\d{4})\-(?P<MM>\d{2})\-(?P<DD>\d{2})' when parsing '2019-03'
arrow.get('2019-03-17', 'YYYY-MM-DD')  # we won this...
# <Arrow [2019-03-17T00:00:00+00:00]>
```

Check the [arrow docs](https://arrow.readthedocs.io/en/latest/) and play around with the format parsing to make sure we don't do anything weird... 

If you find other places in the codebase (even if they're not relevant to this PR), create an issue so we can have a look later.

zenodo/modules/records/serializers/schemas/common.py
```diff
@@ -305,6 +315,26 @@ def validate_language(self, value):
                 field_name=['language']
             )
 
+    @validates('dates')
```
slint: It might be better to move this validation check in the `DateSchemaV1` class. This way, the REST API will return also exactly which items in the array (e.g. `dates[2]`) have an issue, instead of saying that the entire `dates` field has a problem

ChiaraBi: I created this issue: https://github.com/zenodo/zenodo/issues/1749

PR #1748, opened 2019-04-04
Extra formats implementation

Glignos:


zenodo/modules/deposit/utils.py
```diff
@@ -241,3 +242,14 @@ def suggest_language(q, limit=5):
     if lut:
         langs = ([lut, ] + [l for l in langs if l != lut])[:limit]
     return langs
+
+
+@contextmanager
```
slint: This utility function is not needed now, so it can be removed

zenodo/modules/deposit/views.py
```diff
@@ -316,7 +316,7 @@ def to_files_js(deposit):
 
     res = []
 
-    for f in deposit.files:
+    for f in deposit.files:  # FIXME: Refine this...
```
slint: `FIXME` not relevant anymore since we store metafiles in a different bucket

zenodo/modules/records/api.py
```diff
@@ -68,13 +122,35 @@ def __setitem__(self, key, stream):
             self.flush()
 
 
-class ZenodoRecord(Record):
+class MetaFilesIterator(FilesIterator):
```
slint: This class is not needed, since `FilesIterator` does the job

zenodo/modules/records/api.py
```diff
@@ -68,13 +122,35 @@ def __setitem__(self, key, stream):
...
+
+
+class ZenodoRecord(Record, MetaFilesMixin):
     """Zenodo Record."""
 
     file_cls = ZenodoFileObject
 
     files_iter_cls = ZenodoFilesIterator
 
+    metafiles_iter_cls = MetaFilesIterator
```
slint: Same as in `ZenodoDeposit` comment above, this is not necessary

zenodo/modules/records/serializers/derived.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
```
slint: "Derived" metadata sounds too specific to our use-case, i.e. not all metadata are "derived" from their subject. 

I think the namespace (and other places in the code) should be renamed to either `supplementary_metadata` or `metafiles` (or something else?). I like `metafiles`, since it's short and sounds like a "branded" feature. @lnielsen WDYT?

zenodo/modules/records/templates/zenodo_records/box/files.html
```diff
@@ -53,7 +53,7 @@ <h5><i class="fa {{ icon -}}"></i> {{ title -}}</h5>
       </tr>
     </thead>
     <tbody>
-    {%- for file in files|sort(attribute='key') -%}
+    {%- for file in files|sort(attribute='key') if not file.key.startswith(config.ZENODO_METAFILE_KEY_PREFIX) -%}
```
slint: Not needed anymore

zenodo/modules/records/templates/zenodo_records/box/export.html
```diff
@@ -31,4 +31,7 @@ <h4>{{_('Export')}}</h4>
   <li><a href="{{ url_for('invenio_records_ui.recid_export', pid_value=pid.pid_value, format='schemaorg_jsonld') }}">JSON-LD</a></li>
   <li><a href="{{ url_for('invenio_records_ui.recid_export', pid_value=pid.pid_value, format='xm') }}">MARCXML</a></li>
   <li><a href="https://www.mendeley.com/import/?url={{ url_for('invenio_records_ui.recid', pid_value=pid.pid_value, _external=True) }}"><i class="fa fa-external-link"></i> Mendeley</a></li>
+  {%- for mimetype, format_description in (record | metadata_from_record).items() %}
+    <li><a href="{{ url_for('invenio_records_ui.recid_export', pid_value=pid.pid_value, format='me', mimetype=mimetype) }}">{{ format_description }}</a></li>
```
slint: `... format='me' ...` should be changed accordingly to the `config.py`.

requirements.txt
```diff
@@ -121,7 +121,8 @@ invenio-config==1.0.1
 invenio-csl-rest==1.0.0a1
 invenio-db==1.0.3
 invenio-deposit==1.0.0a9
-invenio-files-rest==1.0.0a23.post1
+-e git+https://github.com/slint/invenio-files-rest.git@webpack-post2#egg=invenio-files-rest
```
slint: I think I rebased on a branch I was working... If possible cherry-pick your commits and ignore the `invenio-files-rest`-related commits I did.

zenodo/modules/deposit/views_rest.py
```diff
@@ -53,3 +69,176 @@ def language():
...
+        RecordsBuckets.create(record=record.model, bucket=metadata_bucket)
+    else:
+        metadata_bucket = Bucket.query.get(record['_buckets']['metadata'])
+    return metadata_bucket
+
+
+class DepositMetadataResource(MethodView):
+    """Deposit metadata resource."""
+
+    @validate_metadata_mimetype
```
slint: Mostly relevant for later on, when we refine the permissions, but this decorator should be probably moved down the order, since we first want to make sure the user has the permissions to perform the action, and then reveal validation information

zenodo/modules/deposit/views_rest.py
```diff
@@ -53,3 +69,176 @@ def language():
...
+    return _permission_builder
+
+
+def get_or_create_metadata_bucket(record):
+    if not record['_buckets'].get('metadata'):
+        metadata_bucket = Bucket.create(
+            quota_size=current_app.config['ZENODO_BUCKET_QUOTA_SIZE'],#lignusfix maybe define a smaller limit
+            max_file_size=current_app.config['ZENODO_MAX_FILE_SIZE'],
+        )
+        # locked or unlocked?
```
slint: unlocked

zenodo/modules/deposit/views_rest.py
```diff
@@ -53,3 +69,176 @@ def language():
...
+                abort(403)
+            return f(self, record=record, *args, **kwargs)
+        return decorator
+    return _permission_builder
+
+
+def get_or_create_metadata_bucket(record):
+    if not record['_buckets'].get('metadata'):
+        metadata_bucket = Bucket.create(
+            quota_size=current_app.config['ZENODO_BUCKET_QUOTA_SIZE'],#lignusfix maybe define a smaller limit
```
slint: Yeah, something like 100MB I guess would be enough. We should define a new config variable for this, e.g. `ZENODO_METAFILE_BUCKET_QUOTA_SIZE`

zenodo/modules/deposit/views_rest.py
```diff
@@ -53,3 +69,176 @@ def language():
...
+
+    @validate_metadata_mimetype
+    @pass_record
+    #@check_deposit_permission('read')
+    def get(self, pid, record, **kwargs):
+        """Get metadata."""
+        mimetype = request.headers.get('Content-Type')
+        if mimetype in record.metafiles:
+            fileobj = record.metafiles[mimetype]
+            return fileobj.obj.send_file(trusted=True)  # XXX: DANGER-ZONE
```
slint: Hm, we have to see if this poses any security risks, i.e. do we trust the XML files of the users who can upload them, are there any exploits here? (ping @lnielsen)

zenodo/modules/deposit/views_rest.py
```diff
@@ -53,3 +69,176 @@ def language():
...
+
+class DepositMetadataResource(MethodView):
+    """Deposit metadata resource."""
+
+    @validate_metadata_mimetype
+    @pass_record
+    #@check_deposit_permission('read')
+    def get(self, pid, record, **kwargs):
+        """Get metadata."""
+        mimetype = request.headers.get('Content-Type')
```
slint: In this case, we deduce the MIMEType from the `Accept` header (see `request.accept_mimetype` below)

zenodo/modules/deposit/views_rest.py
```diff
@@ -53,3 +69,176 @@ def language():
...
+    @pass_record
+    #@check_deposit_permission('update')
+    def put(self, pid, record, **kwargs):
+        """Create or replace metadata."""
+        stream, content_length, content_md5 = \
+            current_files_rest.upload_factory()
+        mimetype = request.headers.get('Content-Type')
+        metadata_bucket = get_or_create_metadata_bucket(record)
+        if record.is_published(): # maybe pass record to deposit var to be easier to read
+            pub_recid, published_record = record.fetch_published()
```
slint: For clarity and better readability, maybe proactively renaming some things could help here... E.g. at the beginning of the method do something like:

```python
def put(self, pid, record, **kwargs):
        """Create or replace metadata."""
        depid, deposit = pid, record  # this is actually a deposit endpoint
        ...
```

Then when fetching the published record and pid, reffer to them as `record` and `recid`

zenodo/modules/deposit/views_rest.py
```diff
@@ -53,3 +69,176 @@ def language():
...
+        
+        with db.session.begin_nested():
```
slint: This logic could be extracted into its own `link_record_metabucket(record, bucket)` function

zenodo/modules/deposit/views_rest.py
```diff
@@ -53,3 +69,176 @@ def language():
...
+        return jsonify({
+            'message': 'Metadata for mimetype "{}" deleted.'.format(mimetype)})
+
+    @pass_record
+    #  @check_deposit_permission('read')
+    def options(self, pid, record, **kwargs):
+        """Get a list of all metadata."""
+        metadata_files = (f for f in record.metafiles)
+        # TODO: expose file information...
+        return jsonify({
```
slint: `record.metafiles.dumps()` might be a good response here

zenodo/modules/deposit/views_rest.py
```diff
@@ -53,3 +69,176 @@ def language():
...
+    # NOTE: No permissions (allow all), this is considered metadata (not files)
+    @pass_record
+    def get(self, pid, record, **kwargs):
+        """Get metadata."""
+        # Check querystring first, and then "Accept" header
+        mimetype = (request.args.get('mimetype') or
+                    next((m for m, _ in request.accept_mimetypes), None))
+        if mimetype not in _mimetype_whitelist():
+            abort(400, '"{}" is not an acceptable mimetype.'.format(mimetype))
+        if mimetype in record.metafiles:
```
slint: If `record.metafiles` is `None`, this will probably fail

zenodo/modules/records/api.py
```diff
@@ -52,9 +58,57 @@ def dumps(self):
         return self.data
 
 
-class ZenodoFilesIterator(FilesIterator):
-    """Zenodo files iterator."""
+class MetaFilesMixin(object):
```
slint: This can be renamed to `ZenodoFilesMixin`, since it overrides `files` and adds `metafiles`. Then below `ZenodoRecord`, should inherit from `invenio_records.api.Record` and `ZenodoFilesMixin`

zenodo/modules/deposit/views_rest.py
Glignos: It is never none since if the bucket does not exist the iterator will return an empty array.

zenodo/modules/records/api.py
```diff
@@ -26,14 +26,16 @@
 
 from __future__ import absolute_import
 
+from collections import OrderedDict
 from os.path import splitext
 
-from flask import request
+from flask import current_app, request
```
slint: `current_app` is not used anymore

zenodo/modules/records/serializers/extra_formats.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def serialize(self, pid, record, links_factory=None):
+        """Include files for single record retrievals."""
+        from zenodo.modules.deposit.views_rest import _mimetype_whitelist
+        mimetype = request and request.args.get('mimetype')
+        if not mimetype:
+            return 'Mimetype not specified.'
+        if mimetype not in _mimetype_whitelist():
+            return '"{}" is not an acceptable mimetype.'.format(mimetype)
+        key = mimetype
```
slint: No need for the extra `key` variable here, use `mimetype` directly (possible leftovers from before)

PR #1751, opened 2019-04-05
metadata: add geographical

frankois:


zenodo/modules/deposit/jsonschemas/deposits/records/record-v1.0.0.json
```diff
@@ -649,6 +649,31 @@
...
+          "description": {
+            "description": "Place of the location",
+            "type": "string"
+          },
+          "lat": {
+            "description": "Latitude of the location.",
+            "type": "number"
+          },
+          "lon": {
+            "description": "Lontitude of the location.",
```
ChiaraBi: Small typo: Lontitude -> Longitude

zenodo/modules/deposit/indexer.py
```diff
@@ -98,6 +99,15 @@ def indexer_receiver(sender, json=None, record=None, index=None,
...
             json['relations'] = relations
 
+    try:
+        for index, location in enumerate(json['locations']):
+            lat = location['lat']
+            lon = location['lon']
+            json['locations'][index]['point'] = {'lat': lat, 'lon': lon}
+
+    except:
+        print('No location here')
```
lnielsen: Remove?

zenodo/modules/deposit/indexer.py
```diff
@@ -98,6 +99,15 @@ def indexer_receiver(sender, json=None, record=None, index=None,
         if relations:
             json['relations'] = relations
 
+    try:
+        for index, location in enumerate(json['locations']):
+            lat = location['lat']
+            lon = location['lon']
+            json['locations'][index]['point'] = {'lat': lat, 'lon': lon}
+
+    except:
```
lnielsen: Be very careful with ``except:`` it catches everything including keyboard interrupts. Unless you have very specific reasons for using ``except:`` you should always use ``except Exception:`` instead if you need a generic handle. However, in this case, you should likely go with catching a ``KeyError``.

zenodo/modules/records/indexer.py
```diff
@@ -68,6 +68,15 @@ def indexer_receiver(sender, json=None, record=None, index=None,
         if rels:
             json.setdefault('related_identifiers', []).extend(rels)
 
+    try:
+        for index, location in enumerate(json['locations']):
+            lat = location['lat']
+            lon = location['lon']
+            json['locations'][index]['point'] = {'lat': lat, 'lon': lon}
+
+    except:
```
lnielsen: ditto here

zenodo/modules/deposit/indexer.py
frankois: Thanks, indeed I modified it using ``if json.get('location'):`` instead since the field is not required.

tests/unit/records/test_api_views.py
```diff
@@ -0,0 +1,74 @@
...
+from invenio_indexer.api import RecordIndexer
+
+
+@pytest.mark.parametrize(('val', 'status', 'error_message'), [
+    ("2.45,1.63,-1.43,-1.53", 200, None),
+    ("2.45,1.63", 400,
+     "Invalid bounds: Require 4 numbers separated by comma."),
+    ("2.45,'1.63',-1.43,-1.53", 400, "Invalid number in bounds."),
+    ("2.45,' ',-1.43,-1.53", 400, "Invalid number in bounds."),
+    ("2.45,'',-1.43,-1.53", 400, "Invalid number in bounds."),
```
slint: I would also add a `2.45,,-1.43,-1.53` (and the single-space equivalent `2.45, ,-1.43,-1.53`). Also, spaces around the numbers would be interesting (i.e. maybe we should `.strip()` them): 
```
1.53  ,    -1.43  ,   2.34,1.23
```

Check also other valid values for the `Decimal` type from the docs, like `NaN`: https://docs.python.org/2/library/decimal.html#quick-start-tutorial

PR #1756, opened 2019-04-15
Initial implementation for DCAT serializer

Glignos:


zenodo/config.py
```diff
@@ -703,6 +703,10 @@ def _(x):
         title='Citation',
         serializer='zenodo.modules.records.serializers.citeproc_v1',
     ),
+    'dca': dict(
```
slint:
```suggestion
    'dcat': dict(
```

zenodo/config.py
```diff
@@ -804,6 +808,8 @@ def _(x):
                 'zenodo.modules.records.serializers.dc_v1_response'),
             'application/vnd.citationstyles.csl+json': (
                 'zenodo.modules.records.serializers.csl_v1_response'),
+            'application/dcat+xml': (
```
slint: Is this the recommended `Content-Type`? (e.g. from the spec)

zenodo/config.py
```diff
@@ -1047,6 +1053,12 @@ def _(x):
         'schema': 'http://schema.datacite.org/meta/kernel-3/metadata.xsd',
         'serializer': 'zenodo.modules.records.serializers.oaipmh_oai_datacite',
     },
+    'oai_datacite4': {
```
slint: @lnielsen should we change this to be the default `oai_datacite` format as well?

zenodo/modules/records/serializers/__init__.py
```diff
@@ -77,11 +78,18 @@
 #: DataCite serializer
 datacite_v31 = ZenodoDataCite31Serializer(DataCiteSchemaV1, replace_refs=True)
 datacite_v41 = ZenodoDataCite41Serializer(DataCiteSchemaV4, replace_refs=True)
+#: DCAT serializer
+dcat_v41 = DcatSerializer()
```
slint: Is DCAT's version `4.1`, i.e. does it match DataCite's? If not, `dcat_v1` should be fine

zenodo/modules/records/serializers/dcat.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
```
slint:
```suggestion
# This file is part of Zenodo.
```

Similar mentions below should be updated as well. If this was copy-pasted, check other file(s) as well.

zenodo/modules/records/serializers/dcat.py
```diff
@@ -0,0 +1,58 @@
...
+from lxml import etree as ET
+
+from invenio_oaiserver.response import getrecord
+
+from zenodo.modules.records.serializers.datacite import ZenodoDataCite31Serializer, \
+    ZenodoDataCite41Serializer
+from .schemas.datacite import DataCiteSchemaV1, DataCiteSchemaV4
+
+
+class DcatSerializer(object):
```
slint:
```suggestion
class DCATSerializer(object):
```

zenodo/modules/records/serializers/dcat.py
```diff
@@ -0,0 +1,58 @@
...
+    """CSL Citation Formatter serializer for records"""
+    def serialize(self, pid, record=None, links_factory=None, **kwargs):
```
slint: @lnielsen should there also be a `serialize_search` and `serialize_oaipmh`?

zenodo/modules/records/serializers/dcat.py
```diff
@@ -0,0 +1,62 @@
...
+
+class DCATSerializer(object):
+    """DCAT serializer for records"""
+
+    @cached_property
+    def xslt_transformation(self):
+        current_location = os.path.abspath(os.path.dirname(__file__))
+        file_path = os.path.join(
+            current_location, "../data/datacite-to-dcat-ap.xsl")
+        with open(file_path, 'r') as f:
```
slint: To avoid using relative paths, you can use the `pkg_resources.resource_stream('zenodo.modules.records', 'data/datacite-to-dcat-ap.xsl')` instead

Issue #1695, opened 2019-02-05
records: new field for methods


lnielsen:
Add new text field for recording information about the used Method.

**Data model**

```json
{
   "method": "..."
}
```

- TBD: HTML or plain text?

**Serialization: DataCite XML**

```xml
<descriptions>
<description descriptionType="method">....</date>
</descriptions>
```

**Searlizations**

- JSON
- JSON-LD
- GeoJSON
- MARC21
- ... ?

**Records UI**

- TBD where to display methods.

**Deposit UI**

- Add a new section something like "Datasets" which will hold geographical metata, temporal and method description
- CKEditor if HTML
- Help text explaining what should be put in methods.

**Deposit API:

- Validation: if HTML, then whitelisting.

**Search guide**

Update search guide with new information

**Developer API documentation**

Update developer guide with new information

PR #1757, opened 2019-04-15
deposit: support method field in REST API

ChiaraBi:
* address #1695

### DIFF
--- a/zenodo/modules/records/serializers/schemas/datacite.py
+++ b/zenodo/modules/records/serializers/schemas/datacite.py
@@ -208,6 +208,12 @@ class DataCiteSchema(Schema):
                 })[:max_descr_size],
                 'descriptionType': 'Other'
             })
+        method = obj['metadata'].get('method')
+        if method:
+            items.append({
+                'description': method[:max_descr_size],
+                'descriptionType': 'Methods'
+            })
         return items
 
     def get_rights(self, obj):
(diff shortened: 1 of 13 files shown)


zenodo/modules/records/serializers/schemas/datacite.py
```diff
@@ -422,6 +422,8 @@ class DataCiteSchemaV4(DataCiteSchema):
 
     contributors = fields.Method('get_contributors')
 
+    descriptions = fields.Method('get_descriptions')
```
slint: Since `descriptions` is already defined in the `DataCiteSchema` class, this is not necessary

zenodo/modules/records/serializers/schemas/datacite.py
```diff
@@ -444,6 +446,41 @@ def inject_type(c):
 
         return items
 
+    def get_descriptions(self, obj):
+        """Get descriptions."""
+        items = []
```
slint: To avoid duplication you can call `items = super(DataCiteSchema, self).get_description(obj)`  and then just keep the added `method`-related code and return `items`

zenodo/modules/records/serializers/schemas/datacite.py
```diff
@@ -444,6 +446,41 @@ def inject_type(c):
...
+                        r['raw_reference'] for r in refs
+                        if 'raw_reference' in r]
+                })[:max_descr_size],
+                'descriptionType': 'Other'
+            })
+        method = obj['metadata'].get('method')
+        if method:
+            items.append({
+                'description': method[:max_descr_size],
+                'descriptionType': 'method'
```
slint: I think this should be capitalized, same as the rest of the `descriptionType`:

```suggestion
                'descriptionType': 'Method'
```

PR #1772, opened 2019-04-30
Custom metadata refactor

slint:


tests/unit/records/test_api_views.py
```diff
@@ -84,3 +84,54 @@ def test_geographical_search(es, api, json_headers, record_with_bucket):
...
+    pid, record = record_with_bucket
+    record['custom'] = custom_metadata
+    RecordIndexer().index(record)
+    current_search.flush_and_refresh(index='records')
+    with api.test_request_context():
+        with api.test_client() as client:
+            match_queries = [
+                '[dwc:family]:Felidae',
+            ]
+            for q in match_queries:
```
Glignos: Iteration for one item.

tests/unit/records/test_records_utils.py
```diff
@@ -63,3 +64,24 @@ def test_openaire_type_validation(app):
...
+            ('dwc:family', 'Felidae'),
+            ('dwc:genus', 'Felis'),
+        },
+        custom_text={
+            ('dwc:behavior', 'Plays with yarn, sleeps in cardboard box.'),
+        }
+    )
+
+    result = build_record_custom_fields(record_with_custom_metadata)
+    assert expected == {
```
Glignos: Instead of processing the result, you could have the expected matching it.

tests/unit/records/test_api_views.py
```diff
@@ -84,3 +84,55 @@ def test_geographical_search(es, api, json_headers, record_with_bucket):
...
+                res = client.get(
+                    url_for('invenio_records_rest.recid_list', custom=q),
+                    headers=json_headers)
+                assert len(res.json) == 1
+            res = client.get(
```
Glignos: This test could probably be parametrized as well to avoid repetition.

zenodo/modules/records/__init__.py
```diff
@@ -25,3 +25,10 @@
 """Data model package."""
 
 from __future__ import absolute_import, print_function
+
+from .proxies import current_custom_metadata, current_zenodo_records
+
+__all__ = (
+    'current_custom_metadata',
+    'current_zenodo_records',
```
Glignos: Is this being used somewhere?

PR #1777, opened 2019-05-06
global: change custom field to use arrays

Glignos:


zenodo/modules/records/jsonschemas/records/base-v1.0.0.json
```diff
@@ -588,10 +588,13 @@
     },
     "custom": {
       "additionalProperties": {
-        "type": [
-          "string",
-          "number"
-        ]
+        "additionalProperties": {
```
slint: For defining arrays of string/numbers you need to use the `items` key (see https://json-schema.org/understanding-json-schema/reference/array.html)

zenodo/modules/records/serializers/schemas/common.py
```diff
@@ -459,19 +459,31 @@ def load_custom(self, obj):
...
+            for value in values:
+                if not isinstance(value, term_type):
+                    raise ValidationError(
+                        'Invalid type for term "{0}", should be "{1}".'
+                        .format(term, valid_terms[term]['term_type']),
+                        field_names=['custom'])
+            multiple_values = valid_terms[term]['multiple']
+            if len(values) > 1 and not multiple_values:
+                raise ValidationError(
+                    'Too many values for term "{0}".'
```
slint: Something like `The term "{0}" accepts only one value` is a bit more explicit

PR #1786, opened 2019-05-22
records: change image and video openaire type

ChiaraBi:


zenodo/modules/theme/templates/zenodo_theme/page.html
```diff
@@ -24,6 +24,20 @@
 
 {%- extends "invenio_theme/page.html" %}
 
+{%- block head_meta %}
+<meta charset="utf-8">
```
lnielsen: It's not possible to do something like:
```{{super}}
<meta name="norton-safeweb-site-verification" content="umenay8zh4kswbi568zqp19bqb-jvngusibub1ygib0x3jne9rig0fnmtofm8abb7lkzgltqp5yhm68s5qz4iqqkm39xl2o-p5foixd-1xfq4yig07ugcd1sp5kmyvpe" />
```

PR #1802, opened 2019-07-04
records-ui: fix version box table columns

slint:


slint:
Fixes an issue with some cells of the version box table not having the proper `colspan` value:

Before:
image

After:
image

PR #1830, opened 2019-08-29
global: bump invenio-app and configure flask limiter

Glignos:


tests/unit/conftest.py
```diff
@@ -156,6 +174,14 @@ def default_config(tmp_db_path):
     APP_DEFAULT_SECURE_HEADERS['session_cookie_secure'] = False
 
     return dict(
+        USE_FLASK_LIMITER=False,
+        RATELIMIT_APPLICATION=wrap_rate_limit,
+        RATELIMIT_GUEST_USER='2 per second',
```
slint: You can skip the `USE_FLASK_LIMITER` and `RATELIMIT_*` configuration, and only set them inside the `use_flask_limiter` fixture (just to keep everything in one place)

tests/unit/conftest.py
```diff
@@ -1297,3 +1323,11 @@ def mock_datacite_minting(mocker, app):
         'invenio_pidstore.providers.datacite.DataCiteMDSClient')
     yield datacite_mock
     app.config['DEPOSIT_DATACITE_MINTING_ENABLED'] = orig
+
+
+@pytest.fixture
+def use_flask_limiter(app):
```
slint: I would move this next to the `wrap_rate_limit` function, just to know that these two go together.

PR #1839, opened 2019-09-10
add easy thumbnail view

Glignos:


tests/unit/conftest.py
```diff
@@ -34,6 +34,7 @@
 from copy import deepcopy
 from datetime import date, datetime, timedelta
 from uuid import UUID, uuid4
+import PIL
```
slint: Unused import

tests/unit/conftest.py
```diff
@@ -834,6 +835,21 @@ def record_with_files_creation(db, record_with_bucket):
     return pid, record, record_url
 
 
+@pytest.fixture
+def record_with_image_creation(db, record_with_bucket):
+    """Creation of a full record with files in database."""
+    pid, record = record_with_bucket
+    filename = 'Test.png'
+    with open(os.environ['INVENIO_STATIC_FOLDER'] + '/img/eu.png', 'r') as f:
```
slint: It's better to use the `zenodo` Python package files via `pkg_resource.resource_filename` to get the file

tests/unit/conftest.py
```diff
@@ -834,6 +835,21 @@ def record_with_files_creation(db, record_with_bucket):
...
+        record.files[filename] = f
+    record.files[filename]['type'] = 'png'
+    record.commit()
```
slint: A  `db.session.commit()` is needed here as well

tests/unit/records/test_views.py
```diff
@@ -296,3 +296,22 @@ def test_stats(app, db, minimal_record, stats, expected_result):
         assert expected_result['unique_views'] in template
         assert expected_result['unique_downloads'] in template
         assert 'All versions' not in template
+
+
+def test_record_thumbnail(app, record_with_image_creation):
+    pid, record, record_url = record_with_image_creation
+    for precached_thumbnail in current_app.config['PRECACHED_THUMBNAILS']:
+        with app.test_client() as client:
+            res = client.get(record_url + '/thumbnail/' + precached_thumbnail)
```
slint: Since it's a proper view we can use `flask.url_for` to get the proper URL to use with the test client

zenodo/config.py
```diff
@@ -1403,3 +1409,11 @@ def _(x):
 
 # Error template
 THEME_429_TEMPLATE = "zenodo_errors/429.html"
+
+# Precached thumbnails
+PRECACHED_THUMBNAILS = {
```
slint:
```suggestion
CACHED_THUMBNAILS = {
```

It's kind of like the argument about "pre-reserving" DOIs

zenodo/modules/records/views.py
```diff
@@ -423,3 +426,37 @@ def community_curation(record, user):
 def record_communities():
     """Context processor for community curation for given record."""
     return dict(community_curation=community_curation)
+
+
+def record_thumbnail(pid, record, **kwargs):
```
slint: You can unpack  the `thumbnail` argument here (no need to keep it in `kwargs`)

zenodo/modules/records/views.py
```diff
@@ -423,3 +426,37 @@ def community_curation(record, user):
...
+    """Provide easy access to the thumbnail of a record for predefined sizes.
+
+    We consider the thumbnail of the record as the first image in the files
+    iterator.
+    """
+    if not has_record_perm(current_user, record, 'read-files'):
+        abort(400, 'Bad Request')
```
slint: Check what the return code is for getting a closed-access file for an anonymous user, and replicate this here

zenodo/modules/records/views.py
```diff
@@ -423,3 +426,37 @@ def community_curation(record, user):
...
     return dict(community_curation=community_curation)
+
+
+def record_thumbnail(pid, record, **kwargs):
+    """Provide easy access to the thumbnail of a record for predefined sizes.
+
+    We consider the thumbnail of the record as the first image in the files
+    iterator.
+    """
+    if not has_record_perm(current_user, record, 'read-files'):
```
slint: There's some permission factory for the IIIF endpoint coming from `invenio-iiif` and overridden via config in Zenodo, so I would try to re-use this one

PR #1858, opened 2019-09-18
deposit: add resource_type field in form

Glignos:


zenodo/modules/deposit/views.py
```diff
@@ -263,6 +264,12 @@ def current_openaire_ctx():
     return dict(current_openaire=current_openaire)
 
 
+# @blueprint.app_context_processor
```
slint: This can be removed

zenodo/modules/records/ext.py
```diff
@@ -44,12 +56,36 @@ def __init__(self, app=None):
         if app:
             self.init_app(app)
 
+    @cached_property
+    def resource_types(self):
+        """."""
```
slint: Add a docstring here

zenodo/modules/records/ext.py
```diff
@@ -44,12 +56,36 @@ def __init__(self, app=None):
...
+                continue
+            else:
+                main_type = ''
+            resource_list.append({
+                'type': main_type,
+                'title': item['title']['en'],
+                'id': internal_id
+            })
+        resource_list.sort(key=lambda x: x['title'])
+        resource_list.insert(0, {'title': 'N/A'})
```
slint: Since this is a heavily UI-related mechanic maybe adding the default/null option in the directive might be better

zenodo/modules/records/ext.py
```diff
@@ -26,14 +26,26 @@
...
 
 from . import config
 from .custom_metadata import CustomMetadataAPI
 from .indexer import indexer_receiver
 from .utils import serialize_record, transform_record
 from .views import blueprint, record_communities
+from .proxies import current_zenodo_records
+
+
+def current_zenodo_records_ctx():
```
slint: Since this is a pretty simple function you can add it as a `lambda: dict(current_zenodo_records=current_zenodo_records)` below

zenodo/modules/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -910,6 +926,26 @@
               "urn"
             ],
             "type": "string"
+          },
+          "resource_type": {
+            "additionalProperties": false,
+            "description": "Record resource type.",
+            "properties": {
+              "openaire_subtype": {
```
slint: The `openaire_subtype` field is not needed here

zenodo/modules/records/models.py
```diff
@@ -186,6 +186,12 @@ def _load_data(cls):
                 else:
                     cls.types.add(objtype['internal_id'])
 
+    @classmethod
+    def validate_internal_id(cls, id):
+        """."""
```
slint: Docstring needed (ditto for other places as well)

zenodo/modules/records/models.py
```diff
@@ -214,7 +220,7 @@ def get_types(cls):
     def get_subtypes(cls, type_):
         """Get object type value."""
         cls._load_data()
-        return cls.subtypes[type_]
+        return cls.subtypes.get(type_, [])
```
slint:
```suggestion
        return cls.subtypes.get(type_)
```

I'm a bit hesitant overall about changing the behavior of this. Before if a `type_` was invalid an error would be raised

zenodo/modules/records/serializers/schemas/common.py
```diff
@@ -244,11 +245,39 @@ def validate_data(self, data):
             )
 
 
-class AlternateIdentifierSchemaV1(IdentifierSchemaV1):
+class ResourceType(Schema):
```
slint:
```suggestion
class ResourceTypeMixin(object):
```

Having this as a mixin will help with inheritance problems

Issue #1428, opened 2018-03-01
Provide incentive to Bib(La)TeX, biber, citeproc etc. to support more item types by offering them in .bib export


katrinleinweber:
Currently, several item types are exported as `misc`: https://github.com/zenodo/zenodo/blob/3b8a9d405b8ce593a269beb56c54a552ede894bb/zenodo/modules/records/serializers/bibtex.py#L85-L96. Is that so, because downstream `.bib` processors would treat them as such anyway? If so, how about switching that behaviour and starting to export `@software{…}` for example? Could that help solve this "chicken or egg"-type conundrum and help convince those downstream processors to support new item types that are clearly in demand?

lnielsen:
We switched to ``misc`` because we got quite a lot of support requests complaining about our BibTeX files generating many warnings in downstream tools. I fully agree it would be better to export as ``@software`` or ``@data`` but based on the support requests I don't think many users are ready for it yet.

If you let me know who to convince downstream, I would be happy to be yet another voice.

katrinleinweber:
OK, thanks for the info. Could you please post a list of those tools here? So that interested parties from https://github.com/force11/force11-sciwg/issues/48 could do and fix those. A [biblatex maintainer demo'ed there, that `@software` is basically already usable](https://github.com/force11/force11-sciwg/issues/48#issuecomment-372022915). So, it seems to be a matter of finally updating those downstream tools & styles.

lnielsen:
biblatex is possible, but that's not what most publishers support as I understand. @augustfly can probably comment more on this.

augustfly:
I don't know what to say and I've drafted a response on the force11 thread a couple of times but failed to send it. So I'll respond on that thread instead as the theoretical discussion is better focused there.

augustfly:
It is probably a different thread but I would like you to output `@misc` with the following fields:

- [ ] version
- [ ] publisher (Zenodo)
- [ ] type ([software/dataset...])

we have a [bst style](https://github.com/AASJournals/AASTeX60/blob/master/bst/aasjournal.bst)  almost set up for that (because hacking `@misc` with a custom style file is a lot easier than changing 10,000 authors latex workflows. ).  I'm going to make our next iteration look for the `type` keyword...

katrinleinweber:
[...]

PR #1872, opened 2019-09-23
BibTex enhancements and rate limiting fixes

Glignos:


slint:
This also closes #1428. I think doing this for `dataset` should be fine also

zenodo/modules/records/serializers/bibtex.py
```diff
@@ -90,7 +90,7 @@ def format(self):
             'poster': self._format_misc,
             'presentation': self._format_misc,
             'publication': self._format_publication,
-            'software': self._format_misc,
+            'software': self._format_software,
```
slint: Add `dataset` as well to use the same logic

katrinleinweber:
Thank you :-)

PR #1902, opened 2019-10-31
stats: add Datacite client to provide statistic reports

frankois:


slint:
Included in https://github.com/zenodo/zenodo/pull/2042

PR #1904, opened 2019-11-01
utils: fix local DOI check

Glignos:


zenodo/modules/records/utils.py
```diff
@@ -94,7 +94,7 @@ def serialize_record(record, pid, serializer, module=None, throws=True,
 
 def is_doi_locally_managed(doi_value):
     """Determine if a DOI value is locally managed."""
-    return any(doi_value.startswith(prefix) for prefix in
+    return any(doi_value.startswith(prefix+'/') for prefix in
```
slint:
```suggestion
    return any(doi_value.startswith(prefix + '/') for prefix in
```

zenodo/modules/records/utils.py
```diff
@@ -94,7 +94,7 @@ def serialize_record(record, pid, serializer, module=None, throws=True,
...
                current_app.config['ZENODO_LOCAL_DOI_PREFIXES'])
```
slint: Can you check other places where `ZENODO_LOCAL_DOI_PREFIXES` is used?

PR #1905, opened 2019-11-04
tasks: fix cleanup task key

Glignos:


zenodo/modules/deposit/tasks.py
```diff
@@ -120,9 +120,14 @@ def cleanup_indexed_deposits():
          .query('term', **{'_deposit.status': 'draft'})
          .source(['_deposit.id']))
     res = q.scan()
-    es_depids_info = [(d.to_dict().get('_deposit.id', [None])[0], d.meta.id,
-                       d.meta.index, d.meta.doc_type) for d in res]
-    es_depids = {p[0] for p in es_depids_info}
+    erroneous_depids = es_depids_info = []
```
slint:
```suggestion
    failed_depids, es_depids_info = [], []
```
The word choice is just a nitpick here...

These should be split into two separate statements otherwise you get the following:

```python
In [10]: l1 = l2 = []

In [11]: l1.append('foobar')

In [12]: l1
Out[12]: ['foobar']

In [13]: l2
Out[13]: ['foobar']
```

zenodo/modules/deposit/tasks.py
```diff
@@ -120,9 +120,14 @@ def cleanup_indexed_deposits():
...
+    for d in res:
+        try:
+            es_depids_info.append((d.to_dict()['_deposit']['id'], d.meta.id,
+                                   d.meta.index, d.meta.doc_type))
+            es_depids = {p[0] for p in es_depids_info}
+        except KeyError:
```
slint: Maybe it's better to catch `Exception` here, to cover other unknown failures as well

zenodo/modules/deposit/tasks.py
```diff
@@ -136,3 +141,7 @@ def cleanup_indexed_deposits():
             id=str(deposit_id),
             index=index,
             doc_type=doc_type)
+
+    if erroneous_depids:
+        current_app.logger.warning(
+            'Missing depids from metadata', extra=erroneous_depids)
```
slint: I think `extra` has to be a dictionary (but not sure, check the internets)

zenodo/modules/deposit/tasks.py
Glignos: Missed that.

zenodo/modules/deposit/tasks.py
Glignos: But do we really want to continue the task if we have a new Exception?

slint: Yup, it's about deposits that are deleted, but for some reason exist in the ES index (e.g. because they were postponed for bulk indexing) and are thus visible to users in their "Uploads" view. If they click on them they get a 404 page, and can't really do anything.

PR #1966, opened 2020-04-06
records: landing page admin option to delete user

lnielsen:
* Adds a menu option to delete the a spam user directly from the landing
  page instead of manually having to enter the URL.


slint:
Oh no, I included this in #1967

lnielsen:
Ah, sorry, I just had another batch of annoying spam to delete :-)

slint:
I cherry-picked it actually, because it was a bit more thoughtful :)

PR #1967, opened 2020-04-06
Non-Zenodo DOIs and other fixes

slint:


zenodo/modules/records/templates/zenodo_records/box/edit.html
```diff
@@ -67,6 +67,9 @@
     <li><a href="{{ url_for('zenodo_deposit.delete', pid_value=pid.pid_value) }}" type="submit">
       <span class="text-danger"><i class="fa fa-trash"></i> Delete ...</span>
     </a></li>
+    <li><a href="{{ url_for('zenodo_spam.delete', user_id=record.owners[0]) }}" type="submit">
```
lnielsen, line 70: shouldn't this be ``record._deposit.created_by``?

slint:
E2E tests seem to be not passing... I've tried a couple of remedies, but it seems to be related on how the `live_server` fixture is initialized (something weird happens with the SQLAlchemy DB connection engine not properly closing)

PR #1971, opened 2020-04-13
tokens: initial implementation

slint:


tests/unit/conftest.py
```diff
@@ -521,6 +522,29 @@ def extra_token(app, db, oauth2_client, users):
...
+            expires=datetime.utcnow() + timedelta(hours=10),
+            is_personal=False,
+            is_internal=True,
+            _scopes=tokens_generate_scope.id,
+        )
+        db.session.add(token_)
+    db.session.commit()
+    return dict(
+        token=token_,
+        auth_header=[
```
Glignos, line 542: (Nitpicking)It's not clear to me why you are returning this header here, since you are not using it and it can still be constructed with the returned object.

tests/unit/records/test_permissions.py
```diff
@@ -85,3 +87,52 @@ def test_file_permissions(app, db, record_with_files_creation,
...
+            'iat': datetime.utcnow(),
+            'sub': {'deposit_id': record['_deposit']['id']},
+        },
+        key=rat_generate_token.access_token,
+        algorithm='HS256',
+        headers={'kid': str(rat_generate_token.id)},
+    )
+
+    with app.test_client() as client:
+        file_url = url_for(
```
Glignos, line 112: It would be nice to also add a second record under this user and try to access it with the token to avoid future misbehavior.

PR #2009, opened 2020-06-09
Custom metadata relationships

slint:


tests/unit/records/test_api_views.py
```diff
@@ -91,12 +91,14 @@ def test_geographical_search(es, api, json_headers, record_with_bucket):
...
+     'format: custom=[term]:[value].'),
     ('[dwc:family]', 400, 'The parameter should have the '
-     'format: custom=[field_name]:field_value.'),
+     'format: custom=[term]:[value].'),
     (':Felidae', 400, 'The parameter should have the '
-     'format: custom=[field_name]:field_value.')
+     'format: custom=[term]:[value].')
 ])
 def test_custom_search_validation(es, api, json_headers, record_with_bucket,
                                   val, status, error_message):
```
Glignos, line 104: Nitpicking: I think there is no need to index a record for this test.

tests/unit/records/test_api_views.py
```diff
@@ -118,8 +120,12 @@ def test_custom_search_validation(es, api, json_headers, record_with_bucket,
...
 
 @pytest.mark.parametrize(('query', 'result'), [
-    ('[dwc:family]:Felidae', 1),
-    ('[dwc:family]:foobar', 0)
+    ('[dwc:family]:[Felidae]', 1),
+    ('[dwc:family]:[foobar]', 0),
+    ('[obo:RO_0002453]:[Cat:]', 1),
+    ('[obo:RO_0002453]:[:"Cat flea"]', 1),
+    ('[obo:RO_0002453]:[foobar:]', 0),
+    ('[obo:RO_0002453]:[(foobar OR "Felis catus"):]', 1),
```
Glignos, line 128: Minor: I might be mistaken but it seems to me that there is a test case missing here, for searching a specific relationship, providing both parts of it.

PR #2053, opened 2020-10-02
spam: adds spam prediction when a record is published

Glignos:


requirements.txt
```diff
@@ -250,3 +250,4 @@ WTForms==2.2.1
 WTForms-Alchemy==0.16.9
 WTForms-Components==0.10.3
 zenodo-accessrequests==1.0.0a5
+joblib==0.14.1
```
slint, line 253: I think we also need the `scikit-learn` dependencies for the model to work, otherwise, it would complain with import errors.

zenodo/modules/spam/config.py
```diff
@@ -0,0 +1,53 @@
...
+    user = User.query.get(deposit['_deposit']['owners'][0])
+    user.active = False
+    user.confirmed_at = None
+    delete_user_sessions(user)
+    db.session.commit()
+    send_spam_user_email(user.email)
+    send_spam_admin_email(deposit, user)
+    abort(
+        400,
+        ('This deposit has been automatically marked as spam, please contact '
```
lnielsen, line 44: ``Our spam protection system has classified your upload as a potential spam attempt. As a preventive measure and due to significant increase in spam, we have therefore deactivated your user account and logged you out of Zenodo. Your upload has not been published. If you think this is wrong, please contact our support.``

zenodo/modules/spam/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+
+from flask import current_app
+from celery import shared_task
+import json
+
+from zenodo.modules.spam import current_spam
+from invenio_records.models import RecordMetadata
+
+
+@shared_task(ignore_result=False, max_retries=6, default_retry_delay=10 * 60,
```
lnielsen, line 37: Do we need retries and rate limiting on this request?

zenodo/modules/spam/templates/zenodo_spam/email/spam_user_email.tpl
```diff
@@ -0,0 +1,27 @@
+{#
```
lnielsen, line 1: I would simply put this in the flash message when they publish, and not send an email.

zenodo/modules/spam/ext.py
```diff
@@ -0,0 +1,70 @@
...
+        """Spam detection model."""
+        self._is_cache_loading = True
+        if not current_app.config.get('ZENODO_SPAM_MODEL_LOCATION'):
+            return None
+        return joblib.load(current_app.config['ZENODO_SPAM_MODEL_LOCATION'])
+
+    @property
+    def is_cache_loading(self):
+        """Flag in case another thread is loading already the model."""
+        return getattr(self, '_is_cache_loading', False)
```
slint, line 51: Not sure if this is thread-safe actually, i.e. you might end up with two threads loading the model. You should use `theading.Lock` probably, as [described here](https://stackoverflow.com/a/10181810/2099726).

zenodo/modules/spam/templates/zenodo_spam/email/spam_user_email.tpl
Glignos: The only issue that this is not persistent so then if a user clicks on another tab they will be locked out and if they haven't read the error message they have no awareness of what happened.

zenodo/modules/spam/ext.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, app=None):
+        """Extension initialization."""
+        if app:
+            self.init_app(app)
+
+    def init_app(self, app):
+        """Flask application initialization."""
+        self.app = app
+        self.init_config(app)
+        warm_up_cache.delay()
```
slint, line 62: This is not safe since we don't know if Flask-CeleryExt has been already initialized (i.e. maybe the RabbitMQ connection URL config is not yet loaded/set). An interesting alternative might be [Celery worker lifecycle signals](https://docs.celeryproject.org/en/v4.1.0/userguide/signals.html) (e.g.`worker_init`).

zenodo/modules/spam/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@shared_task(ignore_result=False)
+def check_metadata_for_spam(depid_value, dep_id):
+    if not current_app.config.get('ZENODO_SPAM_MODEL_LOCATION'):
+        return json.dumps(False)
+    deposit = RecordMetadata.query.get(dep_id)
+    is_spam = current_spam.model.predict(
+        [deposit.json['title'] + ' ' + deposit.json['description']])
+    return json.dumps(bool(is_spam[0]))
```
slint, line 44: Does this need to be serialized as JSON? If that's the case I would also go with a more explicit result type (e.g. a dictionary with a `status` key), to avoid implicit boolean checks:

```python
is_spam = check_metadata_for_spam.delay(...).get()
if is_spam:  # this could be None, False, '', 'false', etc.
    ...
```

zenodo/modules/spam/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+from invenio_records.models import RecordMetadata
+
+
+@shared_task(ignore_result=False)
+def check_metadata_for_spam(depid_value, dep_id):
+    if not current_app.config.get('ZENODO_SPAM_MODEL_LOCATION'):
+        return json.dumps(False)
+    deposit = RecordMetadata.query.get(dep_id)
+    is_spam = current_spam.model.predict(
+        [deposit.json['title'] + ' ' + deposit.json['description']])
```
slint, line 43: There might be cases where `title` and `description` are not filled-in (even though they are required fields).

zenodo/modules/deposit/api.py
```diff
@@ -484,8 +487,14 @@ def validate_publish(self):
...
     @mark_as_action
-    def publish(self, pid=None, id_=None, user_id=None, sip_agent=None):
+    def publish(self, pid=None, id_=None, user_id=None, sip_agent=None,
+                check_for_spam=True):
         """Publish the Zenodo deposit."""
+        task = check_metadata_for_spam.delay(self['recid'], str(self.id))
+        is_spam = json.loads(task.get())
+        if is_spam and check_for_spam:
+            if not Permission(ActionNeed('admin-access')).can():
+                current_app.config['ZENODO_SPAM_HANDLING_ACTIONS'](self)
```
slint, line 497: I would wrap this in try/catch with a warning raise in the `catch` block. In case something goes bad in any part of the check, we don't want to completely block users from publishing records.

zenodo/modules/spam/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+from zenodo.modules.spam import current_spam
+from invenio_records.models import RecordMetadata
+
+
+@shared_task(ignore_result=False)
+def check_metadata_for_spam(depid_value, dep_id):
+    if not current_app.config.get('ZENODO_SPAM_MODEL_LOCATION'):
+        return json.dumps(False)
+    deposit = RecordMetadata.query.get(dep_id)
+    is_spam = current_spam.model.predict(
```
slint, line 42: Basing the decision on the float values returned by the `model.predict_proba()` result might give us some better control over the spam threshold... We can then fine-tune the threshold via config if needed.

requirements.txt
Glignos: I thought so too but that is not the case, it seems like scikit is building a standalone python code.

zenodo/modules/spam/ext.py
Glignos: Indeed that looks much cleaner. Maybe this way we can also have a guarantee that the code will run on all processes.

zenodo/modules/spam/tasks.py
Glignos: I am not sure how a record can reach the publication flow without having these 2 fields validated first, maybe we can discuss about these cases IRL.

slint: If one does:

```
# Create deposit
POST /api/deposit/depositions
(no payload)

# Publish
POST /api/deposit/depositions/12345/actions/publish
```

You end up with a record that doesn't have a title or description

zenodo/modules/spam/config.py
```diff
@@ -0,0 +1,56 @@
...
+from zenodo.modules.spam.utils import send_spam_user_email, \
+    send_spam_admin_email
+from flask import abort
+
+
+def default_spam_handling(deposit):
+    """Default actions to counter spam detected record."""
+    user = User.query.get(deposit['_deposit']['owners'][0])
+    user.active = False
+    user.confirmed_at = None
```
slint, line 37: This is unecessary

zenodo/modules/spam/config.py
```diff
@@ -0,0 +1,56 @@
...
+    delete_user_sessions(user)
+    db.session.commit()
+    send_spam_user_email(user.email)
+    send_spam_admin_email(deposit, user)
```
slint, line 41: This could be configurable, to disable in the future if needed

zenodo/modules/spam/tasks.py
Glignos: Yup, I moved it accordingly for the fields to get validated first.

zenodo/modules/deposit/api.py
```diff
@@ -484,10 +488,28 @@ def validate_publish(self):
...
+            else:
+                spam_proba = 0
+            if spam_proba > current_app.config['ZENODO_SPAM_THRESHOLD']:
+                if not Permission(ActionNeed('admin-access')).can():
+                    current_app.config['ZENODO_SPAM_HANDLING_ACTIONS'](self)
+        except TimeoutError:
+            current_app.logger.error(
+                u'Could not check deposit for spam',
+                extra={'depid': self['recid']}
+            )
```
slint, line 511: I think this can go inside a method to keep the `publish` method small

zenodo/modules/spam/config.py
```diff
@@ -0,0 +1,76 @@
...
+    send_spam_user_email(user.email)
+    if current_app.config['ZENODO_SPAM_EMAIL_ADMINS']:
+        send_spam_admin_email(deposit, user)
+    flash(
+        ('Our spam protection system has classified your upload as a '
+         'potential spam attempt. As a preventive measure and due to '
+         'significant increase in spam, we have therefore deactivated your '
+         'user account and logged you out of Zenodo. Your upload has not been '
+         'published. If you think this is a mistake, please contact our '
+         'support.'),
```
slint, line 52: You and store the text message in a variable, and reuse in flash/abort

zenodo/modules/spam/ext.py
```diff
@@ -0,0 +1,82 @@
...
+        for k in dir(config):
+            if k.startswith('ZENODO_SPAM_'):
+                app.config.setdefault(k, getattr(config, k))
+
+
+@celeryd_init.connect
+def warm_up_cache(instance, **kwargs):
+    """Preload the spam model in the celery application."""
+    with instance.app.flask_app.app_context():
+        current_spam.model
```
slint, line 82: This should be inside `init_app`

zenodo/modules/deposit/api.py
```diff
@@ -484,10 +488,28 @@ def validate_publish(self):
...
+                check_for_spam=True):
         """Publish the Zenodo deposit."""
         self['owners'] = self['_deposit']['owners']
         self.validate_publish()
+        try:
+            if check_for_spam and current_app.config.get(
+                    'ZENODO_SPAM_MODEL_LOCATION'):
+                task = check_metadata_for_spam.delay(
+                    self['recid'], str(self.id))
+                spam_proba = task.get(timeout=8)
```
slint, line 501: `timeout` shold be configruable

zenodo/modules/spam/ext.py
Glignos: Not really, this signal gets registered for celery it is not dependent on Flask signals so it doesn't need the app.

slint (approved):

PR #2086, opened 2020-12-17
Fix typo at the front page

kinnala:


slint:
Thanks a lot for your contirbution @kinnala!

PR #2088, opened 2020-12-23
Hot fixes (OpenAIRE link, Github error handler)

Glignos:


Glignos:
Currently, all validation errors are returning a Lazy string that then we fail to commit to the database since it's not JSON serializable. This is causing all failed releases to stay in a `Processing` status not providing any feedback to the users.

PR #2094, opened 2021-01-13
spam: breaking bug fixes

Glignos:


zenodo/modules/spam/utils.py
```diff
@@ -83,15 +84,19 @@ def check_and_handle_spam(community=None, deposit=None):
...
                     Q('query_string', query="owners:{}".format(
                         community.id_user))).count()
                 has_communities = Community.query.filter_by(
-                        id_user=community.id_user).count()
+                        id_user=community.id_user).count() - 1
+
                 if not (has_records or has_communities):
                     current_app.config['ZENODO_SPAM_HANDLING_ACTIONS'](
                         community=community, deposit=deposit)
+    except BadRequest:
```
slint, line 99: Not sure if `werkzeug.exceptions.HTTPException` might be a better choice. E.g. in Zenodo custom errors we sub-class from HTTPException usually.

slint (approved):
Minor thing  about the Exception class, otherwise LGTM

zenodo/modules/spam/utils.py
Glignos: I would avoid this since if for instance the ES query times out we don't want it to be raised.

PR #2098, opened 2021-01-21
deposit-ui: makes the displayed size limit dynamic

Glignos:
* Changes display file size measurement unit to ^10 instead of ^2
* Adds additional support link in the footer


slint (approved):
LGTM, one naming change to protect from name clashes.

PR #2112, opened 2021-02-08
global: invenio-banners integration

slint:


slint:
As part of deployment we also need to run once `zenodo alembic upgrade 5e02314da32e` to create the `banners` table

Glignos (approved):
LGTM

PR #2118, opened 2021-02-19
API Pagination

ctrl-alt-soneca:


tests/unit/records/test_api_views.py
```diff
@@ -141,3 +141,30 @@ def test_custom_search(es, api, json_headers, record_with_bucket,
...
+    record['custom'] = custom_metadata
+    RecordIndexer().index(record)
+    current_search.flush_and_refresh(index='records')
+    with api.test_request_context():
+        with api.test_client() as client:
+            res = client.get(
+                    url_for('invenio_records_rest.recid_list',
+                            all_versions=query),
+                    headers=json_headers)
+            assert res.headers["Link"] == '<http://localhost/' \
```
slint, line 166: Actually, an easier place to check is the response's JSON body (I think via `res.json`), and parse the `links.self` key's value [using `urllib`](https://stackoverflow.com/a/39841923/2099726) to compare.

zenodo/modules/records/query.py
```diff
@@ -45,3 +49,4 @@ def search_factory(self, search, query_parser=None):
     """Search factory."""
     search, urlkwargs = _es_search_factory(self, search)
     return apply_version_filters(search, urlkwargs)
+
```
slint, line 52: An empty line has been added or removed from the end of the file? Normally we prefer having only one empty line at the end of the file.

tests/unit/records/test_api_views.py
```diff
@@ -141,3 +141,30 @@ def test_custom_search(es, api, json_headers, record_with_bucket,
...
                 headers=json_headers)
             assert len(res.json) == result
+
+
+@pytest.mark.parametrize(('query', 'result'), [
+    ("true", "true"),
+    ("True", "true"),
+    ("1", "true"),
+    ("1234", "1234"),
+    ("test", "test"),
```
slint, line 151: I would add a case where the value is `None`, to check the old behavior of `?all_versions`

zenodo/modules/records/query.py
```diff
@@ -30,11 +30,15 @@
 from flask import request
 from invenio_records_rest.query import es_search_factory as _es_search_factory
 
-
 def apply_version_filters(search, urlkwargs):
     """Apply record version filters to search."""
     if request and 'all_versions' in request.values:
-        urlkwargs.add('all_versions', None)
+        if((str(request.values['all_versions']) == "1") or
+                str(request.values['all_versions'].lower()) == "true"):
```
slint, line 37: 1. For readability, `request.values['all_versions']` an be stored in a separate variable (e.g. `all_versions = ...`)
2. `'1'.lower() == '1'`, so the first clause is already covered in the second clause :)
3. The `?all_versions` use-case needs to also be covered (i.e. `all_versions == None`)

Here's how it might look after applying the above:

```suggestion
        all_versions = request.values['all_versions']
        if all_versions is None or str(all_versions).lower() in ("1", "true"):
```

zenodo/modules/records/query.py
```diff
@@ -30,11 +30,15 @@
...
+            urlkwargs.add('all_versions', "true")
+
+        else:
+            urlkwargs.add('all_versions', str(request.values['all_versions']))
```
slint, line 41: Since we assume this is a "falsey" value for the `all_versions` parameter, you have to also apply the search filter below:

```suggestion
            urlkwargs.add('all_versions', str(request.values['all_versions']))
            search = search.filter(Q('term', **{'relations.version.is_last': True}))
```

...or even better store it in a variable and replace the `else` clause below with an `if ...`

slint (commented):
Some extra test cases, styling, and readability. Otherwise LGTM

tests/unit/records/test_api_views.py
ctrl-alt-soneca: In the case where the value is `None` what should be the output?

slint: We treat `None` as "truthy", so it should return `"true"`

ctrl-alt-soneca: Ok and I suppose it goes as `None` and not as `"None"`

tests/unit/records/test_api_views.py
ctrl-alt-soneca: I have changed to this:
```python
parsed_url = urlparse(res.headers["Link"])
parsed_url = parse_qs(parsed_url.query)
assert parsed_url["all_versions"][0] == result
```

The `res.json` outputted `[]`

slint:
Moved to #2138

PR #2135, opened 2021-03-15
cli: fixes missing size update after removing a file

Glignos:


zenodo/modules/utils/cli.py
```diff
@@ -172,6 +172,7 @@ def remove_file(recid, key=None, index=None):
 
     if click.confirm(u'Continue?'):
         bucket.locked = False
+        bucket.size = bucket.size - obj.file.size
```
slint, line 175:
```suggestion
        bucket.size -= obj.file.size
```

slint (approved):
Minor idiomatic/style nit, but not critical

PR #2141, opened 2021-03-18
deposit: edit bucket size option added

ctrl-alt-soneca:


zenodo/modules/deposit/templates/zenodo_deposit/edit.html
```diff
@@ -140,6 +140,35 @@ <h1>Edit upload</h1>
       <h1>New version</h1>
       {%- else %}
       <h1>New upload</h1>
+      {% if record.files %}
+      {%- if current_user_is_admin() %}
```
slint, line 144:
```suggestion
      {# Note: we need an already initialized deposit #}
      {% if record.files and current_user_is_admin() %}
```

More compact, less Jinja blocks nesting, and a comment for some clarity :)

(Remember to remove the `endif` later on. Also test it locally since I can't even remember if that's what Jinja comments looks like :) )

zenodo/modules/deposit/templates/zenodo_deposit/edit.html
```diff
@@ -140,6 +140,35 @@ <h1>Edit upload</h1>
...
+            Record
+          </a></li>
+          <li><a href="{{ url_for('bucket.edit_view', id=record.files.bucket) }}">
+            Bucket
+          </a></li>
+          {%- endif %}
+          <li role="separator" class="divider"></li>
+          <li><a href="{{ url_for('zenodo_deposit.delete', pid_value=pid.pid_value) }}" type="submit">
+            <span class="text-danger"><i class="fa fa-trash"></i> Delete record ...</span>
+          </a></li>
```
slint, line 163: This entry doesn't make a lot of sense in the deposit... Unless you wrap it in an `if is_published` block

zenodo/modules/deposit/templates/zenodo_deposit/edit.html
```diff
@@ -140,6 +140,35 @@ <h1>Edit upload</h1>
...
+          <li><a href="{{ url_for('persistentidentifier.index_view', flt1_21=record.id) }}">
+            Persistent identifiers
+          </a></li>
+          <li><a href="{{ url_for('recordmetadata.details_view', id=record.id) }}">
+            Record
+          </a></li>
+          <li><a href="{{ url_for('bucket.edit_view', id=record.files.bucket) }}">
+            Bucket
+          </a></li>
+          {%- endif %}
```
slint, line 159: Hm, I can't find where the matching `if ` of this block is. Can you double-check?

zenodo/modules/deposit/templates/zenodo_deposit/edit.html
```diff
@@ -140,6 +140,35 @@ <h1>Edit upload</h1>
...
+          Admin <span class="caret"></span>
+        </button>
+        <ul class="dropdown-menu">
+          <li><a href="{{ url_for('persistentidentifier.index_view', flt1_21=record.id) }}">
+            Persistent identifiers
+          </a></li>
+          <li><a href="{{ url_for('recordmetadata.details_view', id=record.id) }}">
+            Record
+          </a></li>
+          <li><a href="{{ url_for('bucket.edit_view', id=record.files.bucket) }}">
```
slint, line 156:
```suggestion
          <li><a href="{{ url_for('bucket.edit_view', id=record.files.bucket.id) }}">
```

It kind of works by accident I feel now (since if you do `print(bucket)` you get [`bucket.id`](https://github.com/inveniosoftware/invenio-files-rest/blob/27e067be62c0b76e96428397cae84555b016f614/invenio_files_rest/models.py#L388))

zenodo/modules/deposit/templates/zenodo_deposit/edit.html
ctrl-alt-soneca: You are right! Since it is not published it does not make sense! Going to remove it!

ctrl-alt-soneca: You are right! Should I add the wrapping? Or should I remove the option?

slint: I think removing is fine. We usually delete records from the public record page (since we're sure it's published then)

PR #2194, opened 2021-07-06
cli: adds list and retry command for not indexed openaire records

Glignos:


zenodo/modules/utils/cli.py
```diff
@@ -407,3 +409,32 @@ def update_openaire_communities(path):
...
                                         separators=(', ', ': '))), fg='blue')
+
+
+@utils.command('list_failed_openaire_records')
+@with_appcontext
+def list_failed_openaire_records():
+    """Lists records that failed to get indexed in OpenAIRE."""
+    failed_record_recids = []
+    for key in current_cache.cache._read_clients.scan_iter(
+        match=(current_cache.cache.key_prefix + 'openaire_direct_index*')):
```
slint, line 421: This iteration logic can be put in an `_iter_openaire_direct_index_keys()` helper function.

zenodo/modules/utils/cli.py
```diff
@@ -407,3 +410,32 @@ def update_openaire_communities(path):
...
+            recid = key.split('openaire_direct_index:')[1]
+            failed_record_recids.append(key)
+    click.echo(failed_record_recids)
```
slint, line 424:
```suggestion
    click.echo('\n'.join(failed_record_recids))
```

So that we get a one-recid-per-line output.

slint (commented):
Some minor changes for clarity, otherwise LGTM

Issue #2127, opened 2021-03-05
deposit-ui: introduce warning pop-up for unpublished deposit "Delete" button


slint:
In a similar fashion to the first "Publish" button warning pop-up, there should be something similar for when users delete unpublished deposits. The warning text should mention:

- files and metadata will be deleted
- that reserved Zenodo DOIs would be lost

**Extra**

- Check why this extra button is there: https://github.com/zenodo/zenodo/blob/master/zenodo/modules/deposit/static/templates/zenodo_deposit/actions.html#L47-L56

PR #2196, opened 2021-07-12
deposit-ui: introduce warning pop-up for unpublished deposit Delete button

ines-cruz:
closes https://github.com/zenodo/zenodo/issues/2127
PRzenodo

### DIFF
--- a/zenodo/modules/deposit/static/templates/zenodo_deposit/actions.html
+++ b/zenodo/modules/deposit/static/templates/zenodo_deposit/actions.html
@@ -3,10 +3,11 @@
     <li class="pull-left">
       <button
         class="btn btn-danger"
+        data-toggle="modal"
+        data-target="#warningDeleteModal"
         ng-hide="recordsVM.invenioRecordsArgs.templateParams.is_published"
         ng-disabled="recordsVM.invenioRecordsLoading || !recordsVM.invenioRecordsEndpoints.self"
-        ng-attr-title="{{(recordsVM.invenioRecordsLoading || !recordsVM.invenioRecordsEndpoints.self) && 'Record was not saved yet.' || 'Delete the deposit'}}"
-        ng-click="recordsVM.actionHandler(['self', 'DELETE'], recordsVM.invenioRecordsArgs.templateParams.is_new_version ? 'latest_html' : '/deposit')">
+        ng-attr-title="{{(recordsVM.invenioRecordsLoading || !recordsVM.invenioRecordsEndpoints.self) && 'Record was not saved yet.' || 'Delete the deposit'}}">
         <i class="fa fa-trash-o fa-inverse"></i> {{ recordsVM.invenioRecordsArgs.templateParams.is_new_version ? 'Discard version' : 'Delete' }}
       </button>
     </li>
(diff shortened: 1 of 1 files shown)


zenodo/modules/deposit/static/templates/zenodo_deposit/actions.html
```diff
@@ -3,10 +3,10 @@
     <li class="pull-left">
       <button
         class="btn btn-danger"
+        data-toggle="modal"
+        data-target="#warningDeleteModal"
         ng-hide="recordsVM.invenioRecordsArgs.templateParams.is_published"
-        ng-disabled="recordsVM.invenioRecordsLoading || !recordsVM.invenioRecordsEndpoints.self"
-        ng-attr-title="{{(recordsVM.invenioRecordsLoading || !recordsVM.invenioRecordsEndpoints.self) && 'Record was not saved yet.' || 'Delete the deposit'}}"
-        ng-click="recordsVM.actionHandler(['self', 'DELETE'], recordsVM.invenioRecordsArgs.templateParams.is_new_version ? 'latest_html' : '/deposit')">
```
ines-cruz, line 9: removed the click part so it wouldn't perform the delete right away but rather show the popup

zenodo/modules/deposit/static/templates/zenodo_deposit/actions.html
```diff
@@ -35,22 +35,7 @@
...
         data-toggle="modal"
         data-target="#warningModal"
-        ng-hide="recordsVM.invenioRecordsArgs.templateParams.is_published"
-        ng-disabled="depositionForm.$invalid || depositionForm.$dirty || recordsVM.invenioRecordsLoading || (filesVM.files | checkAllFilesUploaded) || filesVM.files.length == 0"
-        ng-attr-title="{{(depositionForm.$invalid || depositionForm.$dirty || recordsVM.invenioRecordsLoading) && 'You need to save your deposit first' 
-        || (filesVM.files | checkAllFilesUploaded) && 'There are pending files to be uploaded' 
-        || (filesVM.files.length == 0) && 'You need to upload at least one file' 
-        || 'Publish the record'}}">
-        <i class="fa fa-check"></i> Publish
-      </button>
```
ines-cruz, line 45: There were two Publish buttons, the first one when clicked displayed the popup warning and did not perform the publish part, only showed the popup, the other button was hidden and if clicked it would not show the popup, just perform the action publish so I deleted that one and the publish button now shows the popup and only in the popup there is the action publish

zenodo/modules/deposit/static/templates/zenodo_deposit/actions.html
slint: Hm, just saw you also accidentally removed the `ng-attr-title` attribute, which provides the information mouse hover text. Can you add it back? (`ng-click` can stay removed of course)

zenodo/modules/deposit/static/templates/zenodo_deposit/actions.html
slint: Ditto here for `ng-attr-title`, it should stay.

slint (approved):
Minor changes to add back some attributes for the mouse hover helptext, otherwise LGTM :shipit:

PR #2198, opened 2021-07-15
openaire: virtual access KPIs submission

mitsosf:
Resolves zenodo/ops#95

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -203,6 +203,7 @@ setup(
             'zenodo_theme = zenodo.modules.theme.ext:ZenodoTheme',
             'zenodo_tokens = zenodo.modules.tokens.ext:ResourceAccessTokens',
             'zenodo_spam = zenodo.modules.spam.ext:ZenodoSpam',
+            'zenodo_metrics = zenodo.modules.metrics.ext:ZenodoMetrics',
         ],
         'invenio_base.api_apps': [
             'zenodo_communities = '
@@ -213,6 +214,7 @@ setup(
             'zenodo_exporter = zenodo.modules.exporter.ext:InvenioExporter',
             'zenodo_tokens = zenodo.modules.tokens.ext:ResourceAccessTokens',
             'zenodo_spam = zenodo.modules.spam.ext:ZenodoSpam',
+            'zenodo_metrics = zenodo.modules.metrics.ext:ZenodoMetrics',
         ],
         'invenio_base.blueprints': [
             'zenodo_communities = zenodo.modules.communities.views:blueprint',
@@ -230,6 +232,7 @@ setup(
             'zenodo_rest = zenodo.modules.rest.views:blueprint',
             'zenodo_deposit = zenodo.modules.deposit.views_rest:blueprint',
             'zenodo_spam = zenodo.modules.spam.views:blueprint',
+            'zenodo_metrics = zenodo.modules.metrics.views:blueprint',
         ],
         'invenio_base.api_converters': [
             'file_key = zenodo.modules.deposit.utils:FileKeyConverter',
--- a/zenodo/modules/metrics/api.py
+++ b/zenodo/modules/metrics/api.py
@@ -1,0 +1,89 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Zenodo.
+# Copyright (C) 2016, 2017 CERN.
+#
+# Zenodo is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Zenodo is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Zenodo; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Zenodo Metrics API."""
+
+from __future__ import absolute_import
+
+from elasticsearch_dsl import Search
+from invenio_accounts.models import User
+from invenio_communities.models import Community
+from invenio_files_rest.models import FileInstance
+from invenio_search import current_search_client
+from invenio_search.utils import build_alias_name
+
+from .proxies import current_metrics
+
+
+class ZenodoMetric(object):
+    """API class for Zenodo Metrics."""
+
+    @staticmethod
+    def get_users():
+        """Get number of unique zenodo users."""
+        return 1
+
+    @staticmethod
+    def get_visitors():
+        """Get number of unique zenodo users."""
+        time_range = {'gte': current_metrics.metrics_start_date.isoformat()}
+
+        search = Search(
+            using=current_search_client,
+            index=build_alias_name('events-stats-*')
+        ).filter('range', timestamp=time_range)
+
+        search.aggs.metric(
+            'visitors_count', 'cardinality', field='visitor_id'
+        )
+        result = search.execute()
+
+        if 'visitors_count' not in result.aggregations:
+            return 0
+
+        return result.aggregations.visitors_count.value
+
+    @staticmethod
+    def get_researchers():
+        """Get number of unique zenodo users."""
+        return User.query.filter(
+            User.confirmed_at >= current_metrics.metrics_start_date,
+            User.confirmed_at.isnot(None),
+            User.active.is_(True),
+        ).count()
+
+    @staticmethod
+    def get_files():
+        """Get number of files."""
+        return FileInstance.query.filter(
+            FileInstance.created >= current_metrics.metrics_start_date
+        ).count()
+
+    @staticmethod
+    def get_communities():
+        """Get number of active communities."""
+        return Community.query.filter(
+            Community.created >= current_metrics.metrics_start_date,
+            Community.deleted_at.is_(None)
+        ).count()
--- a/zenodo/modules/metrics/config.py
+++ b/zenodo/modules/metrics/config.py
@@ -1,0 +1,62 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Zenodo.
+# Copyright (C) 2017-2021 CERN.
+#
+# Zenodo is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Zenodo is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Zenodo; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Configuration for Zenodo Metrics."""
+
+from __future__ import absolute_import, print_function
+
+import datetime
+from .api import ZenodoMetric
+
+ZENODO_METRICS_START_DATE = datetime.datetime(2021, 01, 01)
+ZENODO_METRICS_CACHE_TIMEOUT = 3600
+
+ZENODO_METRICS_DATA = {
+    'openaire-nexus': [
+        {
+            'name': 'zenodo_unique_visitors_web_total',
+            'help': 'Number of unique visitors in total on Zenodo portal',
+            'type': 'gauge',
+            'value': ZenodoMetric.get_visitors
+        },
+        {
+            'name': 'zenodo_researchers_total',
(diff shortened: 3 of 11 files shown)


zenodo/modules/metrics/api.py
```diff
@@ -0,0 +1,76 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Zenodo; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Zenodo Communities API."""
```
slint, line 25:
```suggestion
"""Zenodo Metrics API."""
```

zenodo/modules/metrics/api.py
```diff
@@ -0,0 +1,76 @@
...
+from invenio_db import db
+from invenio_pidrelations.contrib.versioning import PIDVersioning
+from invenio_pidstore.models import PersistentIdentifier
+
+
+class ZenodoMetric(object):
+    """API class for Zenodo Metrics."""
+
+    @staticmethod
+    def get_users(cache_timeout=-1):
```
slint, line 38:
```suggestion
    def get_users():
```

I would say the caching can be done on a higher level (i.e. in the `utils.py` file).

zenodo/modules/metrics/config.py
```diff
@@ -0,0 +1,65 @@
...
+api = api.ZenodoMetric()
+
+ZENODO_METRICS = {
+    'openaire-nexus': [
+        {
+            'name': 'kpi.srvs.UsersZenodo',
+            'help': 'Number of unique users of Zenodo',
+            'type': 'gauge',
+            'value': api.get_users(cache_timeout=3600)
+        },
```
slint, line 39:
```suggestion
ZENODO_METRICS = {
    'openaire-nexus': [
        {
            'name': 'kpi.srvs.UsersZenodo',
            'help': 'Number of unique users of Zenodo',
            'type': 'gauge',
            'cache_timeout': 3600,
            'value': ZenodoMetric.get_users,
        },
```

- Since it's a class `staticmethod` you can specify it like this (i.e. no need for an instance)
- The `cache_timeout` can be moved to be a config key

zenodo/modules/metrics/utils.py
```diff
@@ -0,0 +1,46 @@
...
+from flask import Response
+from . import config
+
+
+def openaire_response():
+    response = ''
+    for metric in config.ZENODO_METRICS['openaire-nexus']:
+        response += "# HELP {name} {help}\n# TYPE {name} {type}\n{name} " \
+                    "{value}\n".format(**metric)
+    return Response(response, mimetype='text/plain')
```
slint, line 35:
```suggestion
from flask import current_app

def generate_metrics(metric_id):
    results = []
    for metric in current_app.config['ZENODO_METRICS'][metric_id]:
        metric_value = metric['value']()  # the "caching" logic could be done here also instead
        results.append(**metrics, value=metric_value)
    return results
```

- Config in general, should be fetched from `current_app.config` (since e.g. in production we have a separate config that might override the default)
- Maybe this function could be responsible for generating the results for the metrics, and there's a separate function that does the Prometheus formatting

zenodo/modules/metrics/utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def formatted_response(metric_id):
+    formatters = {
+        'openaire-nexus': openaire_response
+    }
+
+    response = formatters.get(metric_id)
+
+    return Response('Invalid key', status=400, mimetype='text/plain') if \
```
slint, line 45: HTTP formatting can be done in the `views.py`. This function can e.g. just return the formatted text result.

setup.py
```diff
@@ -230,6 +230,7 @@
             'zenodo_rest = zenodo.modules.rest.views:blueprint',
             'zenodo_deposit = zenodo.modules.deposit.views_rest:blueprint',
             'zenodo_spam = zenodo.modules.spam.views:blueprint',
+            'zenodo_metrics = zenodo.modules.metrics.views:blueprint',
```
slint, line 233: Also add under `invenio_base.api_apps` an entry for the extension, i.e. `zenodo.modules.metrics.ext:ZenodoMetrics`

slint (commented):
Some structural changes to keep the view/API logic separated, and some "Flask-isms", but otherwise LGTM

zenodo/modules/metrics/utils.py
mitsosf: I tried using current_app.config, but run into issues (app not being registered properly, still investigating), so I just loaded it this way to see if my core logic worked.

Will refactor it when I manage to register the app properly :)

zenodo/modules/metrics/api.py
mitsosf: The way I envisioned this was that `get_users` would check if there was a cached value and return that or compute it and save it in the cache with `cache_timeout`. This would get repetitive for all functions, but we would eliminate the need for a separate config key.

Works either way for me.

zenodo/modules/metrics/api.py
```diff
@@ -0,0 +1,76 @@
...
+from flask import current_app
+from invenio_accounts.models import User
+from invenio_communities.models import Community
+from invenio_files_rest.models import FileInstance
+
+
+class ZenodoMetric(object):
+    """API class for Zenodo Metrics."""
+
+    metrics_start_date = current_app.config['ZENODO_METRICS_START_DATE']
```
slint, line 38: from werkzeug.locals import LocalProxy

```suggestion
    metrics_start_date = LocalProxy(lambda: current_app.config['ZENODO_METRICS_START_DATE'])
```

That would be one way of making this work. The reason is that `current_app` is bound to what is called the Flask Application Context, which is loaded after some other things have happened first.

So, I agree on fetching this date from configuration, I'm a bit hesitant in having this in such a "global" manner since in the future we might need to expose metrics for different time ranges (and possibly other parameters), in which case we would need a per-metric configuration or similar...

To not overthink things, let's just go with this approach now, and create an issue out of this comment to review in the future.

zenodo/modules/metrics/api.py
```diff
@@ -0,0 +1,76 @@
...
+        return '1'
+
+    @staticmethod
+    def get_researchers():
+        """Get number of unique zenodo users.
+        """
+        return User.query\
+            .filter(User.confirmed_at >= ZenodoMetric.metrics_start_date) \
+            .filter(User.confirmed_at.isnot(None))\
+            .count()
```
slint, line 59:
```suggestion
        return User.query.filter(
            User.confirmed_at >= ZenodoMetric.metrics_start_date,
            User.confirmed_at.isnot(None),
            User.active.is_(True),
        ).count()
```

- In SQLAlchemy for [`User.query.filter(...)`](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query) you can pass multiple clauses as arguments to the function call and they will all be `AND`ed together
- We also need the `User.active.is_(True)`, to filter out deleted/banned/inactivated users.
- For general formatting we don't have it officially configured, but going with something like `black` (I used [this webapp](https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AGXAOJdAD2IimZxl1N_WlXnON2nzOzKrgfhIz2E5t5_Av2bG9X9w2E4-Df9LyfMmZ5zwjoT6e8L4NHaQdkp2XqjN3dyAI6gTFaMc5gTAGv5ipoTEHyjW_tZVceCTIqW21BJoHCrjzS0ZDKb-spI9j_2kY4lX3Y16CtRdQpuxmcdr8khMYS7MOcxGotZvFM-9Klxpz4BBUKIti1RZbfqbL69UWIkcRcdrdHYo5_zXJzodndo5_kmjLjqK_snXOVMpRLLz6_KNtqNqtH2-S1GXx3L5Z9dWpn-dxPK2rRjH-SAeRF51Z80SwAAAABuH1YZEKSlrwAB_gGYAwAAFvIfq7HEZ_sCAAAAAARZWg==)) gives decent results.

zenodo/modules/metrics/api.py
```diff
@@ -0,0 +1,76 @@
...
+            .filter(FileInstance.created >= ZenodoMetric.metrics_start_date) \
+            .count()
+
+    @staticmethod
+    def get_communities():
+        """Get number of unique zenodo users.
+        """
+        return Community.query \
+            .filter(Community.created >= ZenodoMetric.metrics_start_date) \
+            .filter_by(deleted_at=None) \
```
slint, line 75:
```suggestion
            .filter(Community.deleted_at.is_(None)) \
```

FYI, in case you're looking for the `.filter(...)` version of this :)

slint (commented):
Some SQLAlchemy query and formatting fixes.

pytest.ini
```diff
@@ -26,6 +26,6 @@
 pep8ignore =
     __init__.py E501
     docs/conf.py ALL
-addopts = --pep8 --ignore=docs --cov=zenodo --cov-report=term-missing --no-print-logs
+addopts = --pep8 --ignore=docs --no-print-logs
```
slint, line 29: We need the coverage options (also our README.md badge might break).

zenodo/modules/metrics/proxies.py
```diff
@@ -0,0 +1,31 @@
...
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Proxy objects for easier access to application objects."""
+
+from flask import current_app
+from werkzeug.local import LocalProxy
+
+metrics_start_date = LocalProxy(
```
slint, line 30:
```suggestion
current_metrics = LocalProxy(
```

Since this points to the Flask extension

zenodo/modules/metrics/api.py
```diff
@@ -0,0 +1,81 @@
...
+        """Get number of unique zenodo users.
+        """
+        return 1
+
+    @staticmethod
+    def get_researchers():
+        """Get number of unique zenodo users.
+        """
+        return User.query.filter(
+            User.confirmed_at >= ZenodoMetrics().metrics_start_date(),
```
slint, line 61:
```suggestion
            User.confirmed_at >= current_metrics.metrics_start_date,
```

- This would use the already configured/initialized extension (vs instantiating a new one)
- Since `metrics_start_date` is a property, there's no need to make a function call

zenodo/modules/metrics/api.py
```diff
@@ -0,0 +1,81 @@
...
+            User.confirmed_at.isnot(None),
+            User.active.is_(True),
+        ).count()
+
+    @staticmethod
+    def get_files():
+        """Get number of unique zenodo users.
+        """
```
slint, line 69:
```suggestion
        """Get number of files."""
```

zenodo/modules/metrics/api.py
```diff
@@ -0,0 +1,81 @@
...
+        return FileInstance.query.filter(
+            FileInstance.created >= ZenodoMetrics().metrics_start_date()
+        ).count()
+
+    @staticmethod
+    def get_communities():
+        """Get number of unique zenodo users.
+        """
```
slint, line 77:
```suggestion
        """Get number of active communities."""
```

zenodo/modules/metrics/utils.py
```diff
@@ -0,0 +1,57 @@
...
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Utilities for metrics module."""
+from flask import current_app
+from copy import deepcopy
+from invenio_cache import current_cache
+import ipdb
```
slint, line 29:
```suggestion
```

slint (commented):
Reviving this PR :)
Some structural comments to kick it off again. Tests are broken (for other reasons), but let's get the specific code inline to be able to deploy in Sandbox.

tests/unit/conftest.py
```diff
@@ -84,6 +84,7 @@
 from zenodo.modules.records.models import AccessRight
 from zenodo.modules.records.serializers.bibtex import Bibtex
 from zenodo.modules.tokens.scopes import tokens_generate_scope
+from zenodo.modules.metrics.api import ZenodoMetric
```
slint, line 87:
```suggestion
```

Leftover I guess after a move :)

zenodo/modules/metrics/api.py
```diff
@@ -0,0 +1,79 @@
...
+    @staticmethod
+    def get_visitors():
+        """Get number of unique zenodo users."""
+        # Query: '"aggs":{"visitors_count":{"cardinality":{' \
+        #                '"field":"visitor_id","precision_threshold":' \
+        #                '40000}}}'
+        # search = RecordsSearch(index='events-stats-*') \
+        #     .aggs.bucket('visitors_count', 'value_count',
+        #                  field='visitor_id')
+        # return search.count()
```
slint, line 54:
```suggestion
        # Query: '"aggs":{"visitors_count":{"cardinality":{' \
        #                '"field":"visitor_id","precision_threshold":' \
        #                '40000}}}'
        search = Search(
            using=current_search_client,
            index='events-stats-*'
        ).filter(
            'range', ...,
        ).aggs.metric(
            'visitors_count', 'cardinality', field='visitor_id'
        )
        result = search.execute()
        return result.aggregations.visitors_count.value
```

- I think you're right about the `RecordSearch` class, it probably messes things up. You can use the base ElasticSearch-DSL `Search` class as done [here](https://github.com/zenodo/zenodo/blob/9e044a035a4b66d75949c609f29794c44e2b739e/zenodo/modules/stats/exporters.py#L63-L65). Basically you have to pass the `using=<client>` parameter, and apply some index prefixing (i.e. to make `events-stats-*` -> `zenodo-events-stats-*`).
- We also need to filter results in the specific time range (i.e. starting from 2021-01-01), but see the next point.
- For some inspiration for using the `cardinality` metric aggregation over a specific time range see [this example](https://github.com/inveniosoftware/invenio-stats/blob/master/invenio_stats/utils.py#L49-L61).

PR #2205, opened 2021-08-21
records-ui: GitHub branding showing up twice zenodo#2001

### LINKED ISSUE #2001: records-ui: GitHub branding showing up twice
Caused by two of the related identifiers pointing to a GitHub URL.

image

ines-cruz:
closes #2001

I added 3 GitHub repo's to the test recod and only one GitHub branding shows as shown in the pic

askAlex1


ines-cruz:
I have a question regarding this issue. 
I added 3 GitHub repos as supplement of the test record and it showed 3 banners of GitHub, one per repo i added as supplement. Like this, it only has the banner once, for the first supplement repo. I dont think this is a problem because we still have the 3 links to the repos (seen bellow in the pic) but is this how we intend it to be? 
Should that branding of GitHub link to a supplement repo or to the repo of the record?
Thanks!

PR #2215, opened 2021-09-14
deposit: "admin-menu" and "bucket size editing" added

ctrl-alt-soneca:
Captura de ecrã 2021-09-14, às 10 24 56

### DIFF
--- a/zenodo/modules/deposit/templates/zenodo_deposit/edit.html
+++ b/zenodo/modules/deposit/templates/zenodo_deposit/edit.html
@@ -140,6 +140,36 @@
       <h1>New version</h1>
       {%- else %}
       <h1>New upload</h1>
+      {# Note: we need an already initialized deposit #}
+      {% if record.files and current_user_is_admin() %}
+      <div class="btn-group btn-block pull-right">
+        <button type="button" class="btn btn-m btn-default dropdown-toggle pull-right" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+          Admin <span class="caret"></span>
+        </button>
+        <ul class="dropdown-menu">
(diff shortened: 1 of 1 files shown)


zenodo/modules/deposit/templates/zenodo_deposit/edit.html
```diff
@@ -140,6 +140,34 @@ <h1>Edit upload</h1>
...
+          <li><a href="{{ url_for('zenodo_deposit.delete', pid_value=pid.pid_value) }}" type="submit">
+            <span class="text-danger"><i class="fa fa-trash"></i> Delete record ...</span>
+          </a></li>
+          {% if record._deposit.created_by %}
+          <li><a href="{{ url_for('zenodo_spam.delete', user_id=record._deposit.created_by) }}" type="submit">
+            <span class="text-danger"><i class="fa fa-trash"></i> Delete user ...</span>
+          </a></li>
+        </ul>
+      </div>
+      {%- endif %}
```
slint, line 169:
```suggestion
          {% if record._deposit.created_by %}
          <li>
            <a href="{{ url_for('zenodo_spam.delete', user_id=record._deposit.created_by) }}" type="submit">
              <span class="text-danger"><i class="fa fa-trash"></i> Delete user ...</span>
            </a>
          </li>
          {%- endif %}
        </ul>
      </div>
```

Otherwise the `<ul>` wouldn't be closed and might mess up the whole page's HTML

slint (approved):
LGTM, one minor comment for properly closing HTML tags in an edge case.

Last one (if you haven't already), can you check the following scenario to make sure it doesn't break for non-admin users:

1. Non-admin user creates and saves deposit
2. Reload page (check if everything is good)
3. Publish record
4. Edit record (and check that deposit page is good)

slint (approved):

PR #2218, opened 2021-09-21
tests: update APA citation output

slint:


slint:
Closed in favor of #2223

PR #2221, opened 2021-09-23
deposit: allow HTML tables

### LINKED ISSUE #2147: deposit: accept table HTML tags in description
Currently we sanitize HTML input in the `description` field and strip out all tags except for [a curated whitelist](https://github.com/zenodo/zenodo/blob/7327dc80c124cf962f3e66e8b9c48cabeff17536/zenodo/modules/records/serializers/fields/html.py#L33-L55). AFAIK HTML tables do not pose an attack vector, so we can probably add the relevant tags in to the whitelist.

- [x] Update the whitelist with all table-related tags (e.g. `td`, `table`, `th`, etc)
- [x] Check how to enable tables in CKEditor
- [x] Update [Developer docs](https://github.com/zenodo/developers.zenodo.org/) "info" banner with new tags

mitsosf:
closes #2147 

Allows HTML tables to be used in a record's description through the API or CKEditor in the deposit form

### DIFF
--- a/zenodo/modules/records/serializers/fields/html.py
+++ b/zenodo/modules/records/serializers/fields/html.py
@@ -29,7 +29,6 @@ from __future__ import absolute_import, print_function
 import bleach
 
 from .sanitizedunicode import SanitizedUnicode
-
 ALLOWED_TAGS = [
             'a',
             'abbr',
@@ -38,6 +37,7 @@ ALLOWED_TAGS = [
             'blockquote',
             'br',
             'code',
+            'caption',
             'div',
             'em',
             'i',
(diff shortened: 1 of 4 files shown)


mitsosf:
Make sure this is merged as well before closing this

zenodo/developers.zenodo.org#44

zenodo/modules/records/serializers/fields/html.py
```diff
@@ -50,13 +50,27 @@
             'strong',
             'sub',
             'sup',
+            'table',
+            'caption',
```
slint, line 54:
```suggestion
```

Duplicate from above.

zenodo/modules/theme/static/scss/record.scss
```diff
@@ -305,3 +305,27 @@
...
+  table {
+    margin: 1rem 0;
+  }
+
+  th, td {
+    border-style: solid;
+    border-width: thin;
+    padding: 0.5rem 2rem;
+  }
+}
```
slint, line 331:
```suggestion
.record-description, .record-notes {
  table {
    margin: 1rem 0;
  }

  th, td {
    border-style: solid;
    border-width: thin;
    padding: 0.5rem 2rem;
  }
}
```
For avoiding duplication

slint (approved):
Minor changes, but LGTM. I think after addressing the long-text issue we can merge.

slint (approved):

PR #2269, opened 2022-01-02
Fix minor typo in legacy schema

ml-evs:
Can't see anywhere in the code that generated this typo, so I have just fixed it in the schema directly.


slint:
Thank you so much for the contribution @ml-evs :)

PR #2288, opened 2022-01-27
cli: OpenAIRE record deletion tool

ctrl-alt-soneca:


zenodo/modules/utils/cli.py
```diff
@@ -440,3 +445,45 @@ def retry_indexing_failed_openaire_records(eager):
         else:
             openaire_direct_index.delay(
                 str(recid.object_uuid), retry=False)
+
+
+@utils.command('delete_openaire_records')
+@with_appcontext
+def delete_openaire_records():
+    """Deletes half deleted OpenAIRE records."""
+    for key in _iter_openaire_direct_index_keys():
```
slint, line 454: Since a similar operation is done in the `retry_indexing_failed_openaire_records` command above, I would try to merge them. So the flow is something like:

- Loop over `_iter_openaire_direct_index_keys()`
  - Try to fetch the record
  - If the fetched record/PID is deleted, then run this new workflow added in this PR to delete from OpenAIRE
  - If the record is active, do as before (i.e. try to direct index)

zenodo/modules/utils/cli.py
```diff
@@ -440,3 +445,45 @@ def retry_indexing_failed_openaire_records(eager):
...
+                "originalId": original_id, "collectedFromId": datasource_id
+            }
+            req = _openaire_request_factory()
+            res = req.delete(
+                current_app.config["OPENAIRE_API_URL"], params=params
+            )
+            res_beta = None
+            if current_app.config["OPENAIRE_API_URL_BETA"]:
+                res_beta = req.delete(
+                    current_app.config["OPENAIRE_API_URL_BETA"], params=params
```
slint, line 482: Since this logic already exists in the `openaire_delete` task, you can call directly the task here.

zenodo/modules/utils/cli.py
```diff
@@ -440,3 +445,45 @@ def retry_indexing_failed_openaire_records(eager):
...
+            RECID = key.split("openaire_direct_index:")[1]
+
+            record_data = PersistentIdentifier.query.filter_by(
+                pid_type="recid", pid_value=str(RECID), status=PIDStatus.DELETED
+            ).one()
+            record = ZenodoRecord.get_record(
+                    record_data.object_uuid, with_deleted=True
+                    )
+            if "removal_reason" in record:
+                last_valid_revision = record.revisions[-2]
```
slint, line 465: There are some rare cases where the -2th (i.e. previous revision) won't do it, because of past mistakes. A more "natural" way to find the last good revision, is to seek backwards, until e.g. you find a revision that has a `resource_type` field (which is actually what we're looking for to figure out the OpenAIRE IDs).

PR #2302, opened 2022-03-09
deposit_form: update

ctrl-alt-soneca:


zenodo/modules/deposit/static/json/zenodo_deposit/deposit_form.json
```diff
@@ -584,7 +584,7 @@
                     "name": "Nederlandse Organisatie voor Wetenschappelijk Onderzoek (NL)"
                   },
                   {
-                    "value": "10.13039/501100000690",
+                    "value": "10.13039/100014013",
```
slint, line 587: This one is good as is actually, we have to keep the "legacy" name, but also add the new one (once we have the new grants imported for UK Research & Innovation).

slint (approved):

PR #2334, opened 2022-06-03
Switch to SVG version of the CERN logo

avivace:
The CERN logo used in the footer (and in the signup) is considerably low resolution (60x60) and quite noticeable, especially when seen close to the OpenAIRE one, which is 475px high.

Let's switch to an SVG version. I also removed the old png and replaced every reference (footer and signup).

Before:
image

After:
image


slint:
Thanks a lot @avivace for the glorious SVG :)

PR #2341, opened 2022-06-22
spam: search safelist

mitsosf:


zenodo/modules/spam/views.py
```diff
@@ -101,3 +104,36 @@ def delete(user_id):
...
+@blueprint.route('/<int:user_id>/safelist', methods=['POST'])
+@login_required
+def safelist_add_remove(user_id):
+    """Add or remove user from spam safelist."""
+    # Only admin can access this view
+    if not Permission(ActionNeed('admin-access')).can():
+        abort(403)
+
+    user = User.query.get(user_id)
+    if request.form['_method'] == 'post':
```
slint, line 117:
```suggestion
    if request.format['action'] == 'add':
```

Since we don't have any HTTP method tricks, and semantic might read cleaner

zenodo/modules/spam/views.py
```diff
@@ -101,3 +104,36 @@ def delete(user_id):
...
+        # Create safelist entry
+        SafelistEntry.create(user_id=user.id, notes='Added by {} ({})'.format(
```
slint, line 119:
```suggestion
        SafelistEntry.create(user_id=user.id, notes=u'Added by {} ({})'.format(
```
Since we're in Python 2 and you never know :)

zenodo/modules/spam/models.py
```diff
@@ -0,0 +1,96 @@
...
+            entry = cls.query.filter(cls.user_id == user_id).first()
+            db.session.delete(entry)
+            db.session.commit()
+        except:
+            pass
+
+    @classmethod
+    def get_record_status(cls, record):
+        """Get entry by user_id."""
+        for owner_id in record["owners"]:
```
slint, line 93:
```suggestion
        for owner_id in record.get("owners", []):
```

There are some record without owners (legacy data)

zenodo/modules/spam/models.py
```diff
@@ -0,0 +1,96 @@
...
+            if not cls.get_by_user_id(owner_id):
+                return False
+        return True
```
slint, line 96: I would maybe reverse the logic here so that "if **any** of the owners is safelisted, return True"

PR #2347, opened 2022-07-01
safelist: fix ui for records without owners

mitsosf:


zenodo/modules/records/templates/zenodo_records/box/edit.html
```diff
@@ -72,36 +72,38 @@
...
-               value="{{ url_for('invenio_records_ui.recid',
-               pid_value=pid.pid_value) }}"
-        />
-        <button class="btn btn-link" type="submit" {{
-          'disabled' if edit_disabled else ''}}><div style="color: #6aa3d5">
-          <i class="fa fa-pencil-square-o"></i> Add to safelist</div>
-        </button>
-      </form>
-    </li>
+    {% if record._deposit.owners and record._deposit.owners[0] %}
```
mitsosf, line 75: Basically adding this line should fix this

slint:
```suggestion
    {% set has_owners = record._deposit.owners and record._deposit.owners[0] %}
```

Maybe something like this, which you can re-use below to determine if you render the option.

PR #2353, opened 2022-07-15
spam: Safelist admin

mitsosf:


zenodo/modules/spam/views.py
```diff
@@ -105,6 +105,73 @@ def delete(user_id):
...
+        try:
+            owner = user.key
+            if not users.get(owner):
+                users[owner] = \
+                    user_info(
+                        user_id=owner,
+                        last_records=user.records
+                    )
+        except Exception:
+            pass
```
slint, line 167: This is a N+1 type of query in the DB, so it could be optimized into a single query with a [join/subquery eager load](https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html#relationship-loading-with-loader-options):

```python
db.session.query(User).options(
    db.joinedload(User.profile, User.external_identifiers)
)
```

zenodo/modules/spam/views.py
```diff
@@ -105,6 +105,87 @@ def delete(user_id):
...
+        # TODO maybe make it more beautiful?
+        try:
+            if user.external_identifiers:
+                for identifier in user.external_identifiers:
+                    info["external_identifiers"] = (
+                        info["external_identifiers"] + identifier.method + ", "
+                    )
+                info["external_identifiers"] = info["external_identifiers"][
+                    :-2
+                ]
```
slint, line 137: I would try to pass more structured data and then delegate the display in the template.

```suggestion
            if user.external_identifiers:
                info["external_identifiers"] = [i.method for i in user.external_identifiers]
```

zenodo/modules/spam/views.py
```diff
@@ -105,6 +105,87 @@ def delete(user_id):
...
+        except Exception:
+            pass
+
+        last_records = owners[user.id]['last_records']
+        if last_records:
+            for record in last_records:
+                info["last_records"] = (
+                    info["last_records"] + record.title + ", "
```
slint, line 145: There are some records with very long titles which might cause some display issues, so we might get some funny renderings.

zenodo/modules/spam/views.py
```diff
@@ -105,6 +105,87 @@ def delete(user_id):
         ctx.update(records=records)
         return render_template('zenodo_spam/delete.html', **ctx)
 
+def user_info(owners):
```
slint, line 108:
```suggestion
def expand_users_info(users):
```

zenodo/modules/spam/templates/zenodo_spam/safelist/admin.html
```diff
@@ -0,0 +1,275 @@
...
+      }
+
+      let userIds = []
+      data.forEach(row => {
+        userIds.push(row[1])
+      })
+
+      $.post( "{{ url_for('zenodo_spam.safelist_bulk_add') }}",
+        {'user_ids': userIds, 'next': {{ url_for('zenodo_spam.safelist_admin')
+        }} })
```
slint, line 258:
```suggestion
        {'user_ids': userIds})
```

`next` and redirects are not that necessary here, since we're making an AJAX call.

zenodo/modules/spam/views.py
```diff
@@ -138,3 +219,30 @@ def safelist_add_remove(user_id):
...
     return redirect(request.form['next'])
+
+@blueprint.route('/safelist/add/bulk', methods=['POST'])
+@login_required
+def safelist_bulk_add():
+    """Add uses to the safelist in bulk."""
+    # Only admin can access this view
+    if not Permission(ActionNeed('admin-access')).can():
+        abort(403)
+    for user_id in request.post['user_ids']:
```
slint, line 230:
```suggestion
    for user_id in request.form['user_ids[]']:
```

I think jQuery does some weird thing here with the naming... At least what I remember from other places we use jQuery :)

mitsosf: This was mostly a WIP, as I didn't get a chance to test this e2e, will change if needed

zenodo/modules/spam/templates/zenodo_spam/safelist/admin.html
mitsosf: Once again, WIP from the copy pasting, indeed we can handle it on the frontend

zenodo/modules/spam/views.py
mitsosf: I thought about cutting the title short, eg, display the first 20 chars of each title, will update it

zenodo/modules/spam/views.py
mitsosf: Yup, had to made that change and then some :)

PR #2361, opened 2022-08-09
spam: safelist admin panel optimizations

mitsosf:
- [x] Dashboard operations trigger too many Celery tasks
    - [x] Commit to database partial information about the action (e.g. inactivate user when marking spam)
    - [x] Take into account database information when filtering ElasticSearch results (e.g. don't show in the table users that were already safelisted or inactivated)
- [x] Bugfix: Calling `SafelistEntry.create(...)` on already safelisted users during bulk safelisting fails with a 500
    - Added another try-except, but seems weird to me since it's already supposed to be handled in the method
- [x] Switch spam delete and reindexing tasks to the `low` queue
- [x] Be able to filter by date ranges
    - Filter by weeks
- [x] List also number of total users using ES unique count aggregation (since we now only list the top 1000 users)
    - Went around this by adding a field to select how many users we want to be displayed
- [x] UI/UX
    - [x] When clicking the user ID to get the user admin spam view, it should open a new tab by default
    - [x] Improve table width by shortening some text values for columns
    - [x] Add record "description" text as well to help with search filtering the table
- [x] Add translation links for record titles
    - Descriptions don't seem to work well for this, maybe a copy button could work better
- [x] Prevent selection of row while clicking on the whole row
    - Selection is triggered when eg. we click on the user profile or translate links

### DIFF
--- a/zenodo/modules/spam/templates/zenodo_spam/safelist/admin.html
+++ b/zenodo/modules/spam/templates/zenodo_spam/safelist/admin.html
@@ -30,17 +30,35 @@
 <div class="container record-admin">
   <div class="row">
     <h1>Safelist admin</h1>
-    <form>
+    <form class="form-inline">
       <div class="form-group">
         <div class="input-group">
-          <input type="number" min="1" class="form-control" name="weeks" placeholder="Weeks (default: 4)">
+          <input type="number" min="1" class="form-control" name="from_weeks"
+                 placeholder="From Weeks ago(4)" value="{{ request.args.get
+                 ('from_weeks') }}">
         </div>
       </div>
+      <div class="form-group">
+        <div class="input-group">
+          <input type="number" min="0" class="form-control" name="to_weeks"
+                 placeholder="To Weeks ago (0)" value="{{ request.args.get
+                 ('to_weeks') }}">
+        </div>
+      </div>
+       <div class="form-group">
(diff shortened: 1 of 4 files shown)


zenodo/modules/spam/templates/zenodo_spam/safelist/admin.html
```diff
@@ -49,17 +67,30 @@ <h1>Safelist admin</h1>
...
-              <td>{{ user.last_records }}</td>
+              <td>
+                <a target="_blank" href="{{ url_for
+                ('zenodo_spam.delete', user_id=user.id) }}">{{ user.id }}</a>
+              </td>
+              <td class="selectable">{{ user.username }}</td>
+              <td class="selectable">{{ user.email }}</td>
+              <td class="selectable">{{ user.full_name }}</td>
+              <td class="selectable">{{ user.external | join(', ') }}</td>
+              <td class="selectable">{{ user.last_records }}</td>
```
lnielsen, line 85: Comment: Can you make a link to the record itself? That would save me quite some time in determining if a user is spammer.

lnielsen (approved):

ppanero (approved):
Would the joins in the db cause a bit of strain? but in any case is "just" the admin panel so it does not have an effect on the end user.

PR #2376, opened 2022-08-24
spam: add domain validation in safelist admin panel

mitsosf:
Adding status flags on the safelisting panel

image


tests/unit/conftest.py
```diff
@@ -241,7 +251,9 @@ def default_config(tmp_db_path, tmp_spam_domains_file):
             },
         },
         SEARCH_INDEX_PREFIX='zenodo-test-',
-        ZENODO_SPAM_DOMAINS_FILEPATH=tmp_spam_domains_file,
+        ZENODO_SPAM_DOMAINS_FORBIDDEN_LIST_FILEPATH
+        =tmp_spam_domains_forbidden_list_file,
```
slint, line 255: Hm, I think this one will probably get a complaint from pep8

zenodo/modules/spam/safelist.py
```diff
@@ -0,0 +1,39 @@
...
+# along with Zenodo; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+
+class DomainSafelist:
```
slint, line 26: To avoid duplication maybe what we need is just data-structure-like class, with a `.matches()` method in `utils.py` which can then be reused like:

```python
# in ext.py to initialize the lists from config
forbidden = DomainList(...)
safelisted = DomainList(...)

# when checking domains
forbidden.matches(email_domain)
safelisted.matches(email_domain)
```

We can then get rid of `forbidden.py` and `safelist.py`

zenodo/modules/spam/config.py
```diff
@@ -93,5 +93,8 @@ def default_spam_handling(deposit=None, community=None):
 # Preload spam model on Celery app initialization
 ZENODO_SPAM_MODEL_PRELOAD = False
 
-ZENODO_SPAM_DOMAINS_FILEPATH = None
+ZENODO_SPAM_DOMAINS_FORBIDDEN_LIST_FILEPATH = None
 """Path to a file with the list of forbidden email domains."""
+
+ZENODO_SPAM_DOMAINS_SAFELIST_FILEPATH = None
+"""Path to a file with the list of safelisted domains."""
```
slint, line 100:
```suggestion
ZENODO_SPAM_DOMAINS_FORBIDDEN_PATH = None
"""Path to a file with the list of forbidden email domains."""

ZENODO_SPAM_DOMAINS_SAFELIST_PATH = None
"""Path to a file with the list of safelisted domains."""
```

Since we're renaming, let's go for something shorter too :)

zenodo/modules/spam/views.py
```diff
@@ -121,6 +123,30 @@ def delete(user_id):
...
+
+    _, domain = _parse_email(email)
+
+    if current_domain_forbidden_list.is_forbidden(domain):
+        return 'forbidden'
+
+    if current_domain_safelist.is_safelisted(domain):
+        return 'safe'
+
+    return 'unclear'
```
slint, line 147:
```suggestion
def _evaluate_user_domain(email):
    _, domain = _parse_email(email)
    if current_domain_forbidden_list.is_forbidden(domain):
        return 'forbidden'
    if current_domain_safelist.is_safelisted(domain):
        return 'safe'
    return 'unclear'
```

zenodo/modules/spam/views.py
```diff
@@ -146,6 +172,7 @@ def _expand_users_info(results):
             r.update({
                 "full_name": user.profile.full_name,
                 "username": user.profile.username,
+                "flag": _evaluate_user_domain(user.email)
```
slint, line 175: This can go in the previous `r.update(...)` since all users have an `.email` (but not a `user.profile`)

tests/unit/spam/test_safelist.py
```diff
@@ -0,0 +1,40 @@
...
+"""Test safelist."""
+
+import os
+import pytest
+import tempfile
+
+from zenodo.modules.spam.safelist import DomainSafelist
+
+
+def test_forbidden_list_search(tmp_spam_domains_safelist_file):
```
slint, line 34: Given the other changes proposed for the overall class structure, I would just keep one test for the utility class itself. 

I don't see any "E2E" style tests anywhere that directly use `current_domain_forbidden_list` and `current_domain_safelist`, in which case I'm not sure if the `conftest.py` additions are necessary also.

slint (commented):
LGTM overall, just some refactoring to avoid duplicating tests/classes

tests/unit/conftest.py
mitsosf: Got this autoformatted from my IDE, just tried with Black online and it seems a one-liner seems to work indeed

zenodo/modules/deposit/utils.py
```diff
@@ -265,7 +265,10 @@ def is_user_verified():
     """Permission function that evaluates if the user can create a deposit."""
     if current_user.email:
         email_domain = current_user.email.rsplit('@', 1)[-1].lower()
-        if current_domain_forbidden_list.is_forbidden(email_domain):
+        forbiddenlist = DomainList(current_app.config.get(
```
slint, line 268: We need this to be initialized in `ext.py` and be exposed via the proxy, since then the file is only loaded once (at application startup), and not every time this check happens.

zenodo/modules/spam/views.py
```diff
@@ -121,6 +122,32 @@ def delete(user_id):
...
+def _parse_email(email):
+    try:
+        prefix, domain = email.split('@', 1)
+        prefix = prefix.lower().strip()
+        domain = domain.lower().strip()
+        if domain[-1] == ".":
+            domain = domain[:-1]
+        return prefix, domain
+    except ValueError:
+        return None, None
```
slint, line 134: Can you check if this kind of logic is used in other places as well, and re-use this function? I think I've seen this before :)

mitsosf: I got it from one of Lars' scripts, but we are using a one-liner elsewhere, so I swapped it :)

slint (approved):

PR #2380, opened 2022-09-15
spam: add domain check for communities creation

mitsosf:


zenodo/modules/communities/utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+from __future__ import absolute_import, unicode_literals
+
+def can_create_community():
+    """Checks it the user can create community."""
+    current_date = datetime.now()
+    confirmed_date = getattr(current_user, 'confirmed_at', None)
+    # TODO Add this variable in config
+    community_user_confirmed_since = \
+        current_app.config['COMMUNITIES_USER_CONFIRMED_SINCE']
```
slint, line 35: This check can go, so that we can have a unified set of "rules" as with the deposit check in `is_user_verified`

slint (approved):

PR #2391, opened 2022-10-18
record: add Horizon Europe (HE) Funder program

mitsosf:


mitsosf:
Not sure if this the right strings or where to find them. I checked the latest dump and it seems like this is contained there.

zenodo/modules/records/config.py
```diff
@@ -160,3 +160,9 @@
 
 
 """
+
+ZENODO_FUNDER_PROGRAM_MATCHES = {
```
slint, line 164:
```suggestion
ZENODO_RECORDS_FUNDER_PROGRAM_MATCHES = {
```

With the `ZENODO_RECORDS_` prefix you'll get it automatically loaded also via the extension (this solves the issue of providing a default below).

zenodo/modules/records/serializers/schemas/datacite.py
```diff
@@ -539,7 +534,14 @@ def get_fundingreferences(self, obj):
...
             award_program = g.get('program')
             if award_program:
-                for funder_id, regexes in self.FUNDER_PROGRAM_MATCHES.items():
+                funder_program_matches = current_app.config.get(
+                    'ZENODO_FUNDER_PROGRAM_MATCHES', {
+                        '10.13039/100010661': ('^H2020$', '^Horizon 2020',),
+                        '10.13039/100011102': ('^FP7$',),
+                        '10.13039/100018693': ('^HE$', '^Horizon Europe',),
+                    }
+                )
```
slint, line 543:
```suggestion
                    'ZENODO_FUNDER_PROGRAM_MATCHES', {})
```

We can leave it empty by default, since there's always going to be something from default config.

mitsosf:
Before merging, let's make sure the strings are also correct.

slint (approved):
The matching text for Horizon Europe is good, so I'm merging :)

PR #2398, opened 2022-11-02
safelist: add filter to see records or/and communities

mitsosf:


zenodo/modules/spam/views.py
```diff
@@ -173,71 +173,80 @@ def safelist_admin():
     if not Permission(ActionNeed('admin-access')).can():
         abort(403)
 
+    data = request.args.get('data', 'all', type=str)
+    data_categories = {
+        'records': True if data == 'all' or data == 'records' else False,
+        'communities': \
+            True if data == 'all' or data == 'communities' else False
```
slint, line 180:
```suggestion
        'records': data in ('all', 'records'),
        'communities': data in ('all', 'communities'),
```

slint (commented):
One minor comment to clarify some boolean checks, otherwise LGTM.

You'll need to rebase this on top of `master` as well (and there's going to be a conflict). Check https://github.com/zenodo/zenodo/commit/543739a8ffa61cc68724fc9ad6898e4d20b7e864 and https://github.com/zenodo/zenodo/commit/e68f4682001246c3b88219d30e837c003598ab3e.

slint (approved):

PR #2403, opened 2022-11-08
safelist: gmail accounts limitation

ctrl-alt-soneca:


zenodo/modules/spam/views.py
```diff
@@ -125,17 +126,34 @@ def delete(user_id):
         return render_template('zenodo_spam/delete.html', **ctx)
 
 
-def _evaluate_user_domain(email):
-    email_domain = email.rsplit('@', 1)[-1].lower()
+def normalize_emails(email):
```
slint, line 129:
```suggestion
def normalize_email(email):
```
Since it's only one

zenodo/modules/spam/views.py
```diff
@@ -125,17 +126,34 @@ def delete(user_id):
...
+    email = email.lower()
+    username = email[: email.index("@")]
+    no_dots_username = username.replace(".", "")
+    domain = email[email.index("@") :]
```
slint, line 133: Let's normalize the domain as well l, i.e. googlemail.com should become gmail.com

zenodo/modules/spam/views.py
```diff
@@ -125,17 +126,34 @@ def delete(user_id):
...
+    return email[email.index("@") :].lower()
+
+
+def _evaluate_user_domain(email, normalized_emails_list):
+    email_domain = email.rsplit("@", 1)[-1].lower()
     if current_domain_forbiddenlist.matches(email_domain):
-        return 'forbidden'
+        return "forbidden"
+    if (get_domain(email) == "@gmail.com") and (
+        normalized_emails_list.get(normalize_emails(email)) > 3
```
slint, line 146: For clarity you can break down these boolean checks into variables.

zenodo/modules/spam/views.py
```diff
@@ -144,8 +162,13 @@ def _expand_users_info(results, include_pending=False):
         ).filter(User.id.in_(results.keys()))
     )
 
+    normalized_emails_list = Counter([
```
slint, line 165:
```suggestion
    normalized_emails = Counter([
```
Since it's not a list 😏

zenodo/modules/spam/views.py
```diff
@@ -144,8 +162,13 @@ def _expand_users_info(results, include_pending=False):
...
+        normalize_emails(user.email)
+        for user in user_data
+    ])
+
     for user in user_data:
-        if user.safelist or (not user.active and not include_pending):
+        if (user.safelist or not user.active) and not include_pending:
```
slint, line 171: This slightly changes the filtering, but looking at it again I see that it might have been wrong all along. Let's discuss IRL on how it should be.

slint (commented):
LGTM, some minor changes and a discussion to be made

zenodo/modules/spam/views.py
```diff
@@ -126,17 +127,38 @@ def delete(user_id):
         return render_template('zenodo_spam/delete.html', **ctx)
 
 
-def _evaluate_user_domain(email):
-    email_domain = email.rsplit('@', 1)[-1].lower()
+def normalize_email(email):
+    email = email.lower()
+    username = email[: email.index("@")]
```
slint, line 132:
```suggestion
    username, domain = email.rsplit("@", 1)
```

This makes sure that if someone comes with `alex@slint@gmail.com` we get the domain correctly.

Note that this way you'll get `username == "alex@slint"` and `domain == "gmail.com"` (without the `@` sign), so you'll have to add the `@` later on.

zenodo/modules/spam/views.py
```diff
@@ -126,17 +127,38 @@ def delete(user_id):
...
+    no_dots_username = username.replace(".", "")
+    domain = email[email.index("@") :]
+
+    if(domain == "@googlemail.com"):
```
slint, line 136:
```suggestion
    if domain == "@googlemail.com":
```

No need for parentheses here :)

zenodo/modules/spam/views.py
```diff
@@ -126,17 +127,38 @@ def delete(user_id):
...
+    return no_dots_username + domain
+
+
+def get_domain(email):
+    return email[email.index("@") :].lower()
+
+
+def _evaluate_user_domain(email, normalized_emails):
+    is_flagged_domain = (get_domain(email) == "@gmail.com")
+    is_above_threshold = (normalized_emails.get(normalize_email(email)) > 3)
```
slint, line 147:
```suggestion
    is_flagged_domain = get_domain(email) == "@gmail.com"
    is_above_threshold = normalized_emails.get(normalize_email(email), 0) > 3
```

- Removed parentheses
- Provide `0` as a default value, to avoid the comparison of `None > 3` (which would fail)

slint (commented):
Couple of small changes and we're good.

zenodo/modules/spam/views.py
```diff
@@ -126,13 +127,33 @@ def delete(user_id):
         return render_template('zenodo_spam/delete.html', **ctx)
 
 
-def _evaluate_user_domain(email):
-    email_domain = email.rsplit('@', 1)[-1].lower()
+def normalize_email(email):
+    email = email.lower()
+    username, domain = email.rsplit("@", 1)
+    no_dots_username = username.replace(".", "")
+    domain = email[email.index("@") :]
```
slint, line 134: minor: why not reuse the `domain` variable from the split above?

ctrl-alt-soneca: Thank you for the suggestion :)

PR #2405, opened 2022-11-09
permissions: fix rat token check

mitsosf:


zenodo/modules/records/permissions.py
```diff
@@ -376,12 +377,15 @@ def has_read_files_permission(user, record):
...
-        if rat_deposit_id == deposit_id and rat_access == 'read':
-            return has_update_permission(rat_signer, record)
+        try:
+            rat_signer, payload = decode_rat(rat_token)
+            rat_deposit_id = payload.get('deposit_id')
+            rat_access = payload.get('access')
+            deposit_id = record.get('_deposit', {}).get('id')
+            if rat_deposit_id == deposit_id and rat_access == 'read':
+                return has_update_permission(rat_signer, record)
+        except MissingTokenIDError:
```
slint, line 387: minor: I would add a comment here, to explain why we do this. Something like "because the `token` querystring param is also used for secret link tokens, they might be misinterpeted as RAT tokens and validation."

slint (commented):
LGTM, maybe a small comment to help anyone reading this in the future

mitsosf:
Added comment and rebased.

PR #2408, opened 2022-11-28
Safelist filtering fixes

mitsosf:
Still missing a corner case and testing, but this is shipable.

### DIFF
--- a/zenodo/modules/spam/models.py
+++ b/zenodo/modules/spam/models.py
@@ -65,14 +65,19 @@ class SafelistEntry(db.Model):
         try:
             entry = cls(user_id=user_id, notes=notes)
             db.session.add(entry)
-            return entry
         except Exception:
-            return None
+            entry = None
+
+        return entry
 
     @classmethod
     def get_by_user_id(cls, user_id):
         """Get entry by user_id."""
-        entry = cls.query.filter(cls.user_id == user_id).first()
+        try:
+            entry = cls.query.filter(cls.user_id == user_id).first()
+        except Exception:
+            entry = None
+
         return entry
 
     @classmethod
--- a/zenodo/modules/spam/views.py
+++ b/zenodo/modules/spam/views.py
@@ -52,6 +52,7 @@ from zenodo.modules.spam.models import SafelistEntry
 from zenodo.modules.spam.proxies import current_domain_forbiddenlist, \
     current_domain_safelist
 from zenodo.modules.spam.tasks import delete_spam_user, reindex_user_records
+from zenodo.modules.spam.utils import is_user_safelisted
 
 blueprint = Blueprint(
     'zenodo_spam',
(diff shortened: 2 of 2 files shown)


mitsosf:
I believe we covered all corner cases.

zenodo/modules/spam/views.py
```diff
@@ -248,10 +251,13 @@ def safelist_admin():
             sa.func.max(Community.description).label('description')
         ).join(Community).group_by(User.id).filter(
             Community.created.between(from_date, to_date),
-            Community.deleted_at == None,
+            Community.deleted_at is None
         ).limit(max_users)
 
         for row in community_users:
+            if is_user_safelisted(row.user_id):
+                continue
```
slint, line 259:
```suggestion
            Community.deleted_at is None,
            ~User.safelist.any(),
        ).limit(max_users)

        for row in community_users:
```

To avoid extra `N` queries after the big one :)

There's a JOIN-variant for this, but not sure if t changes much performance-wise.

zenodo/modules/spam/views.py
```diff
@@ -248,10 +251,13 @@ def safelist_admin():
             sa.func.max(Community.description).label('description')
         ).join(Community).group_by(User.id).filter(
             Community.created.between(from_date, to_date),
-            Community.deleted_at == None,
+            Community.deleted_at is None
```
slint, line 254:
```suggestion
            Community.deleted_at.is_(None),
```

This is a weird behaviour from SQLAlchemy... The one you had resolves to `False`, which causes some weird behaviour in the produced SQL.

zenodo/modules/spam/views.py
```diff
@@ -155,7 +156,7 @@ def _evaluate_user_domain(email, normalized_emails):
     return "unclear"
 
 
-def _expand_users_info(results, include_pending=False):
+def _expand_users_info(results, include_pending=True):
```
slint, line 159:
```suggestion
def _expand_users_info(results, include_pending=False):
```

I think by default we want to not include pending (which is anyways controlled by the querystring argument in the view)

slint (commented):
Some query improvements/fixes, and default behaviour, and we're good to :ship:

zenodo/modules/spam/models.py
```diff
@@ -59,20 +59,27 @@ class SafelistEntry(db.Model):
...
     def create(cls, user_id, notes=None):
         """Create a safelist entry."""
         try:
-            entry = cls(user_id=user_id, notes=notes)
-            db.session.add(entry)
-            return entry
+            existing_entry = cls.get_by_user_id(user_id=user_id)
+            if not existing_entry:
+                entry = cls(user_id=user_id, notes=notes)
+                db.session.add(entry)
```
slint, line 79: minor: Instead of baking this check in the `.create`, I would check in the view `if user.safelist:`, to avoid checking/fetching twice.

zenodo/modules/spam/views.py
mitsosf: I thought it was okay since we're doing the iteration anyway, but this way there are fewer things to iterate through indeed.

slint (approved):

PR #2410, opened 2022-11-29
openaire: indexing retry improvements

slint:


mitsosf (approved):

zenodo/config.py
```diff
@@ -291,7 +291,11 @@ def _(x):
     'github-tokens-refresh': {
         'task': 'invenio_github.tasks.refresh_accounts',
         'schedule': crontab(minute=0, hour=3),
-    }
+    },
+    'openaire-failures-retry': {
```
ppanero, line 295: moderate: should this go to the low priority queue? If yes, should we play with the time delta to avoid having yet another bunch of tasks at the same time (thus triggering notifications?)

zenodo/modules/openaire/tasks.py
```diff
@@ -134,5 +143,30 @@ def openaire_delete(record_uuid=None, original_id=None, datasource_id=None):
 
         if not res.ok or (res_beta and not res_beta.ok):
             raise OpenAIRERequestError(res.text)
+
+        # Remove from failures cache
+        current_cache.delete(
+            'openaire_direct_index:{}'.format(record['recid']))
```
ppanero, line 149: lack of knowledge question: how/when are the failed tasks registered under this key in the cache?

ppanero (approved):
2 minor questions, mostly for my knowledge. Nice work!

zenodo/config.py
slint: - OpenAIRE direct indexing is actually equivalent to DOI registration in terms of importance, so regular priority is correct in this case.
- Regarding timing, I put every 24h (without a specific time), since there's no specific requirement (it's similar to sitemap generation). But having a specific time (maybe even during working hours) might be better, I'll change it to 09:00am UTC.

zenodo/modules/openaire/tasks.py
slint: Nice catch, you're right, failed "deletions" are not registered... [Direct indexing sets failed recids](https://github.com/zenodo/zenodo/blob/0059ad30e73b229de5b8394f5f73b4020dbf7b11/zenodo/modules/openaire/tasks.py#L99) though. I will add it for the deletions as well.

PR #2411, opened 2022-11-29
dependencies: bump invenio-opendefinition

ppanero:
- closes https://github.com/zenodo/ops/issues/251


ppanero:
Discussed IRL, due to depencency rabbit hole we will simply override the handler in Zenodo's conifg. Then we will properly bump/migrate in RDM.

PR #2413, opened 2022-11-30
spam: add more info on delete pages

mitsosf:
image

### DIFF
--- a/zenodo/modules/records/templates/zenodo_records/macros/admin.html
+++ b/zenodo/modules/records/templates/zenodo_records/macros/admin.html
@@ -39,6 +39,25 @@
   <strong>Active?:</strong> {{user.profile and user.active}}
 </div>
 <div class="list-group-item col-sm-6">
+  <strong>Linked accounts:</strong>
+  {% for ext_id in user.external_identifiers %}
+  {% if ext_id.method == 'orcid' %}
+    <img class="inline-orcid"
+         src="{{ url_for('static', filename='img/orcid.png')}}"
+    />
+    <a target="_blank"
+       href="{{ext_id.id | pid_url(scheme='orcid', url_scheme='https')}}">
+      {{ ext_id.id }}
+    </a>
+  {% elif ext_id.method == 'github' %}
+      <i class="fa fa-github"></i>
+    <a target="_blank" href="https://api.github.com/user/{{ext_id.id}}">
+      {{ ext_id.id }}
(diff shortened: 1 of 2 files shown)


zenodo/modules/records/templates/zenodo_records/macros/admin.html
```diff
@@ -67,7 +73,12 @@
 {%- endmacro %}
 
 {%- macro safelist_action(user, action) %}
-  {{user is safelisted_user}}
+  {% if user is safelisted_user %}
```
slint, line 76: As discussed IRL, if you apply the [`uselist=False` parameter](https://github.com/inveniosoftware/invenio-accounts/blob/8d993b7f6239891e6d645a3857ba5ca492849974/invenio_accounts/models.py#L137) on the `db.relationship` of the model, you could instead do something like:

```suggestion
  {% if user.safelist %}
```

mitsosf: Don't remember why we decided to go with this, ofc it works, but a reminder would be appreciated :D

zenodo/modules/records/templates/zenodo_records/macros/admin.html
```diff
@@ -67,7 +91,14 @@
 {%- endmacro %}
 
 {%- macro safelist_action(user, action) %}
-  {{user is safelisted_user}}
+  {% if user.safelist %}
+    {% set safelister = user | safelisted_by %}
+    {% set safelisted_on = user.safelist[0].created.date() %}
```
slint, line 96:
```suggestion
    {% set safelisted_on = user.safelist.created.date() %}
```

Piggy-backing on https://github.com/zenodo/zenodo/pull/2413#discussion_r1071272100, this would be possible if you do something [here](https://github.com/zenodo/zenodo/blob/b87a52f58679976d38a53c93e2425999e469e4f7/zenodo/modules/spam/models.py#L60) like:

```python
    user = db.relationship(User, backref='safelist', uselist=False)
```

slint (approved):
LGTM, one clarification for the `uselist=False`

zenodo/modules/records/templates/zenodo_records/macros/admin.html
mitsosf: This didn't work for me for some reason, let's discuss IRL when you have a chance.

PR #2419, opened 2022-12-05
record: request deletion button

mitsosf:
image

image

### DIFF
--- a/zenodo/modules/records/templates/zenodo_records/box/edit.html
+++ b/zenodo/modules/records/templates/zenodo_records/box/edit.html
@@ -45,6 +45,53 @@
     </form>
   {%- endif %}
 {%- endif %}
+{%- if current_user|has_record_perm(record, 'update') %}
+  {% set eligible_for_deletion = record.created is not older_than(days=30) %}
+  {% if eligible_for_deletion %}
+    <button class="btn btn-danger btn-block" data-toggle="modal"
+            id="request-record-deletion"
+            data-target="#request-delete-modal">
+      <i class="fa fa-trash"></i> Request record deletion
+    </button>
+
+    <div class="modal fade" id="request-delete-modal" tabindex="-1" role="dialog" aria-labelledby="request-delete-modal" aria-hidden="true">
+      <div class="modal-dialog" role="document">
+        <div class="modal-content">
+          <div class="modal-header">
+            <h5 class="modal-title" id="requestDeletionModalLabel">Are you sure
+              you want to request deletion of this record?</h5>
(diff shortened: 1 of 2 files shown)


zenodo/modules/records/templates/zenodo_records/box/edit.html
```diff
@@ -109,3 +154,45 @@
   </ul>
 </div>
 {%- endif %}
+
+<script>
```
slint, line 158: Ideally, with the changes discussed IRL, we can completely remove JS since:

- there's no UI validation for the DOI typing
- the action is stored using cookies that expire in 1 week

zenodo/modules/records/views.py
```diff
@@ -559,3 +566,64 @@ def record_extra_formats(pid, record, mimetype=None, **kwargs):
...
+        return redirect(url_for("invenio_records_ui.recid", pid_value=record[
+        'recid']))
+
+    # Validate form data
+    if request.form['expected-doi'] != request.form['record-doi']:
+        flash(_("The DOI you typed did not match the record's DOI."),
+              category='warning')
+        return redirect(url_for("invenio_records_ui.recid", pid_value=record[
+        'recid']))
+
```
slint, line 588: Not necessary as discussed IRL

zenodo/modules/records/templates/zenodo_records/box/edit.html
```diff
@@ -29,6 +29,51 @@
 <form action="{{ url_for('zenodo_deposit.edit', pid_value=pid.pid_value) }}" method="POST">
   <button class="btn btn-warning btn-block" type="submit" {{ 'disabled' if edit_disabled else ''}}><i class="fa fa-pencil-square-o"></i> Edit</button>
 </form>
+  {% set eligible_for_deletion = record.created is not older_than(days=30) %}
```
slint, line 32: Regarding placement, as discussed IRL should go under the New Version button

zenodo/modules/records/templates/zenodo_records/box/edit.html
mitsosf: I am doing the cookie check with JS, because if I didn't I would have to modify the landing page view to check for the cookie and it seems excessive to me, let's discuss IRL, maybe I'm forgetting parts of our last discussion.

mitsosf:
Before this goes live we need to see how we will be handling DOIs not managed by us, eg provide a perpetual option for deletion. Or simplify things code-wise by having the users go through the support line if the record is older than 30 days?

PR #2422, opened 2022-12-12
github: remove version and publication date from .cff file

mitsosf:


slint (approved):
It might be that some of the tests fail, in which case we just have to update the produced/asserted metadata :)

PR #2428, opened 2023-01-18
spam: add domain count in safelist

mitsosf:
Adds another column with total users domain count in the safelisting panel

### DIFF
--- a/zenodo/modules/spam/views.py
+++ b/zenodo/modules/spam/views.py
@@ -143,18 +143,28 @@ def get_domain(email):
     return email[email.index("@") :].lower()
 
 
-def _evaluate_user_domain(email, normalized_emails):
-    is_flagged_domain = get_domain(email) == "@gmail.com"
+def _evaluate_user_domain(email, normalized_emails, email_domain_count):
+    email_domain = get_domain(email)
+    is_flagged_domain = email_domain in ["@gmail.com"]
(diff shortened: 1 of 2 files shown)


zenodo/modules/spam/views.py
```diff
@@ -172,6 +182,14 @@ def _expand_users_info(results, include_pending=False):
         for user in user_data
     ])
 
+    email_domains = db.session.query(
+        sa.func.split_part(sa.func.lower(User.email),'@',2),
+        sa.func.count(User.id),
+        sa.func.count(User.id).filter(User.active == True))\
+        .group_by(sa.func.split_part(sa.func.lower(User.email),'@',2)).all()
+    email_domain_count = {domain: {'active': active, 'total': total} for
+                          domain, active, total in email_domains}
```
slint, line 191: Not a big fan on the formatting here, but I know I'm risking an aggressive AI takeover by saying that :)

I would also breakdown some parts of the query for readability:

```suggestion
    email_domain_expr = sa.func.split_part(sa.func.lower(User.email),'@',2)
    email_domains = db.session.query(
        email_domain_expr,
        sa.func.count(User.id),
        sa.func.count(User.id).filter(User.active == True)
    ).group_by(email_domain_expr).all()
    
    email_domain_count = {
        domain: {'active': active, 'total': total} 
        for domain, active, total in email_domains
    }
```

PR #2444, opened 2023-03-14
datacite: removing dactite 3

cenouralm:


zenodo/config.py
```diff
@@ -1265,11 +1265,6 @@ def safelist_sort(_):
         'schema': 'http://schema.datacite.org/meta/kernel-4.1/metadata.xsd',
         'serializer': 'zenodo.modules.records.serializers.oaipmh_datacite_v41',
     },
-    'datacite3': {
```
slint, line 1268: I missed on below `oai_datacite3`, and the `oai_datacite should have the same config as `oai_datacite4`

zenodo/config.py
```diff
@@ -935,7 +935,7 @@ def _(x):
             'application/x-bibtex': (
```
slint, line 935: Search also for `ZenodoDataCite31Serializer`, `DataCite31Serializer`, and `DataCiteSchemaV1` which should all be removed. This will probably break tests also (so you might have to remove some of the tests as well.

slint (commented):
There are a couple of references left to remove.

PR #2446, opened 2023-03-16
search: fix mappings for records and deposits

mitsosf:


ppanero:
Nice! much needed cleanup, thanks! I am wondering if some cases would need both (i.e. text + keyword). For example name, if I search for `pab` I would like `pablo` to be returned. If it is keyword only `pablo` (i.e. exact match) would return a hit.

mitsosf:
Yes, we will take a final look at the end to discuss what changes stay or go, just doing a first pass.

slint (commented):
Some minor corrections for description-like fields and a special case. Check also that they're applied on the `record`  mapping.

slint:
> For example name, if I search for `pab` I would like `pablo` to be returned. If it is keyword only `pablo` (i.e. exact match) would return a hit.

I tried to verify this and had to change again my entire internal mental model of how ES types work...  So, here's the summary:

<table>
<thead>
  <tr>
    <th>Value 👉 </th>
    <th colspan="2">"Pablo Panero"</th>
    <th colspan="2">"Panero"</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>Query 👇 | Type 👉 </td>
    <td><code>text</code></td>
    <td><code>keyword</code></td>
    <td><code>text</code></td>
    <td><code>keyword</code></td>
  </tr>
  <tr>
    <td><code>pan</code></td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td><code>panero</code></td>
    <td>✅</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td><code>pablo</code></td>
    <td>✅</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td><code>pablo*</code></td>
    <td>✅</td>
    <td>✅</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td><code>pan*</code></td>
    <td>✅</td>
    <td>✅</td>
    <td>✅</td>
    <td>✅</td>
  </tr>
</tbody>
</table>

- Partial matches (i.e. just `pan`) don't work... Word stemming doesn't actually apply by default. Wildcards (i.e. `*`) and regex queries work though :)
- Case sensitivity doesn't matter in any case... I think it's because [the standard analyzer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-analyzer.html) anyways lower-cases everything

The result of this is that we probably need both in some cases, e.g. author names (actually that would solve a big problem of trying to find all records by `Panero, Pablo`, without also getting results where [`Panero, Jose`, `Garcia, Pablo`] are authors).

zenodo/modules/deposit/mappings/v7/deposits/records/record-v1.0.0.json
```diff
@@ -222,22 +222,22 @@
         "type": "date"
       },
       "title": {
-        "type": "text"
+        "type": "keyword"
```
slint, line 225: Based on https://github.com/zenodo/zenodo/pull/2446#issuecomment-1479311413, I would say we keep this as `text`

zenodo/modules/deposit/mappings/v7/deposits/records/record-v1.0.0.json
```diff
@@ -359,7 +359,7 @@
       "references": {
         "properties": {
           "raw_reference": {
-            "type": "text"
+            "type": "keyword"
```
slint, line 362: Keep `text`, and apply the same `.keyword` trick as above.

zenodo/modules/deposit/mappings/v7/deposits/records/record-v1.0.0.json
```diff
@@ -371,7 +371,7 @@
             "type": "keyword"
           },
           "title": {
-            "type": "text"
+            "type": "keyword"
```
slint, line 374: Same `text` + `fields.keyword` trick.

slint (commented):
Some reconsiderations for existing fields, and applying the `fields.keyword` trick for others. Afterwards, we're good and should be able to merge.

PR #2452, opened 2023-04-17
permissions: allow list for public buckets

slint:
* Allows listing files of public buckets, like e.g. the exporter
  bucket.


lnielsen:
Should this be the case? For exporter I think it's fine, but this would also allow communities logos no?

